PR #1, opened 2019-01-22
Migrate to mono-repo for better command dependency isolation

amcgee:
This PR:

- Makes the repository a monorepo with separate packages for each top-level CLI module
- Extracts the CLI utility functions (mainly `namespace`, `makeEntryPoint`, `cache`, and `reporter`) into a separate repository for easy re-use across CLI modules
- Extracts the `create-d2-app` into its own package (to be published as `@dhis2/create-app`)
- Ports [@dhis2/packages](https://github.com/dhis2/packages) into a CLI module `d2 repo`
- Dogfoods `d2-repo` for monorepo management (some issues with this below)
- Renames `backend` to `cluster`

Simple usage:

```sh
> git clone https://github.com/dhis2/cli
> cd cli
> npx @dhis2/packages link # THIS IS A STOPGAP UNTIL WE HAVE d2-repo PUBLISHED
> npx @dhis2/packages install # THIS IS A STOPGAP UNTIL WE HAVE d2-repo PUBLISHED
# Note the extra directories with node_modules created in packages now
> ./packages/d2/bin/d2
> ./packages/d2/bin/d2 cluster init # cache the docker-compose configuration
> ./packages/d2/bin/d2 cluster up # this will download and spin up a full DHIS2 docker cluster
> ./packages/d2/bin/d2 cluster status
> ./packages/d2/bin/d2 cluster down # destroy the created containers
> ./packages/d2/bin/d2 debug cache status # inspect the cache at ~/.cache/d2
> ./packages/d2/bin/d2 debug cache purge # get rid of the globally-cached docker-compose directory
> ./packages/d2-app/bin/d2-app create
> ./packages/d2-repo/bin/d2-repo
# ...etc
# You could also link d2 globally with:
> cd ./packages/d2 && yarn link
> d2 # profit
```

ISSUES WITH d2-repo:
- `d2 repo link` generates a build directory with modified package.json.  This makes sense in some special cases like our frontend libs, but doesn't make sense in this case.  We don't even need a build script yet, and when we do it will build packaged executables, the published package should still be a node module.
- `d2 repo link` (which was previously `packages link`) creates directories like `packages/app` for each CLI modules package (`d2-app`, `d2-cluster`, `d2-repo`) with `node_modules/@dhis2/d2-cli` linked dependency.  This should be created in `packages/d2-app` not `packages/app`, I don't know where it's getting the name it's using.

Before diving into fixing these bugs, **should we just use Lerna in this case and keep `d2 repo` specialized for frontend d2 app/library mono-repos?**

A few additional questions:
1. If `d2 repo` is special, maybe it should be namespaced under `d2 app repo` instead of a top-level module?  Code-style and commit-style could live in the top-level namespace if we want.  I like consolidation but also functional isolation.
2. `d2 repo link` doesn't appear to be linking executables other than `d2` at the top-level `node_modules/.bin`.  I imagine this is because `@dhis2/cli` is the only package published so far, so it doesn't link in executables from unpublished packages?
3. Do we like `d2 repo` as a name?  I'm not sure, but best I've come up with.
3. I'm proposing publishing each CLI module as `@dhis2/cli-<name>` i.e. `@dhis2/cli-app`.  This leaves `@dhis2/cli-utils` as the odd-man-out.  It works for now, but is a bit confusing and would prevent us from ever having a `d2 utils` namespace.  Thoughts on package naming?

NOTE: I made minimal changes to the code from the `packages` repo, so there's likely some duplication with things we could take from `@dhis2/cli-utils` or elsewhere in the monorepo.


varl:
>     * `d2 repo link` generates a build directory with modified package.json.  This makes sense in some special cases like our frontend libs, but doesn't make sense in this case.  We don't even need a build script yet, and when we do it will build packaged executables, the published package should still be a node module.

Right. The use-case in this repo matches what Lerna does nicely so using the specialized "flatpacking" tool here might not make enough sense to enable our flatpacker to deal with this "non-built" source.

>     * `d2 repo link` (which was previously `packages link`) creates directories like `packages/app` for each CLI modules package (`d2-app`, `d2-cluster`, `d2-repo`) with `node_modules/@dhis2/d2-cli` linked dependency.  This should be created in `packages/d2-app` not `packages/app`, I don't know where it's getting the name it's using.

The naming convention assumed by _the tool formerly known as `packages`_ was that you have e.g. `d2` at the top-level, and the sub-packages would get named `d2-subpkg-name`. E.g. for **libraries**, for example `d2-ui`:

- top-level package name is `@dhis2/d2-ui`
- path for subpackage: `packages/core`
- generated subpackage name: `@dhis2/d2-ui-core`

For applications it works out as:

- top-level package name is `data-visualizer`
- path for subpackage: `packages/app` & `packages/plugin`
- generated subpackage name: `data-visualizer-app` & `data-visualizer-plugin`

We can fix this through removal of the code which identifies the conventions, or moving the structure here to drop the `d2-` prefix, e.g. `packages/app`, `packages/repo`. If we want to do the former, now is a good time to do so.

> Before diving into fixing these bugs, **should we just use Lerna in this case and keep `d2 repo` specialized for frontend d2 app/library mono-repos?**

I think it is sensible to use Lerna here. It was designed for Node monorepos and works exceedingly well there. _But_ I strongly advise using the centrally managed version system to make sure that all packages are published with the same version _especially_ since we have a `utils/` package.

I wrote the `packages` tool when it became clear that Lerna is _never_ going to support it (after talking to the dev), and reading the codebase to see if I could fork Lerna and add it myself. The codebase in Lerna is fantastic, but it would require a ton of work to have it do what we need as it is designed to _not_ do that in the first place.

> A few additional questions:
> 
>     1. If `d2 repo` is special, maybe it should be namespaced under `d2 app repo` instead of a top-level module?  Code-style and commit-style could live in the top-level namespace if we want.  I like consolidation but also functional isolation.

Are we using the `app` cmd/name-space for libs as well? I think a lot of the same applies, so maybe we can then alias `d2 lib` to `d2 app` until we need to split them apart?

I agree with functional isolation as a principle to drive the structure, it becomes relatively obvious that we need to "open it up" if we start seeing things like `d2 app repo link` inside of packages which are obviously not apps (if we treat decide to treat `d2-repo` as a special tool).

>     2. `d2 repo link` doesn't appear to be linking executables other than `d2` at the top-level `node_modules/.bin`.  I imagine this is because `@dhis2/cli` is the only package published so far, so it doesn't link in executables from unpublished packages?

Under the hood the `d2 repo link` cmd uses symlinks. I changed this implementation away from `yarn link` when we started seeing a ton of problems for developers because `yarn link` creates and maintains the links in user space, not isolated to the repo itself.

This also means that the mechanism to link the binaries has to be coded in the `link` command of `d2-repo`.

>     3. Do we like `d2 repo` as a name?  I'm not sure, but best I've come up with.

I don't think it explains what the code actually does, it doesn't _really_ care about the fact that it's a repository, it cares about _packages_, hence the original name. It doesn't do anything related to Git repos except help you stage files when you run `release`.

`code-style`, or `commit-style` are technically separate but do the same.

`d2 packages` with alias to `pkgs` maybe?

>     4. I'm proposing publishing each CLI module as `@dhis2/cli-<name>` i.e. `@dhis2/cli-app`.  This leaves `@dhis2/cli-utils` as the odd-man-out.  It works for now, but is a bit confusing and would prevent us from ever having a `d2 utils` namespace.  Thoughts on package naming?

We could take a hint from Babel's playbook and call our subpackage `helper-utils` to not squat the `d2-utils` namespace, which gives us a complete namespace to use for internal packages: `@dhis2/cli-helper-utils`, `@dhis2/cli-helper-something`. I think it's better to squat the `d2 helper` namespace than to squat the `d2 utils` namespace. We might want to use `d2-utils` but I'm not seeing `d2-helper` being necessary.

README.md
```diff
@@ -27,9 +27,10 @@ You can also run the CLI ad-hoc with `npx`, no installation necessary (sacrifice
 d2 <command>
 
 Commands:
-  d2 app      Manage DHIS2 applications                             [aliases: a]
   d2 debug    Debug local d2 installation
+  d2 app      Manage DHIS2 applications                             [aliases: a]
   d2 cluster  Manage DHIS2 Docker clusters                          [aliases: c]
+  d2 repo     Manage DHIS2 repositories                             [aliases: r]
```
varl: As noted in my long comment, an alternative name is `d2 packages` (alias `d2 pkgs`).

amcgee:
> See long comment for discussion.
> 
> One question from my side:
> 
> I don't see the code for `dhis2/packages` anywhere, and no links referring to it in `package.json` files. Is the plan to merge the `dhis2/packages` repo into this or to have `dhis2/packages` and just mod it to work as a pluggable command in its own repo?

The code from [dhis2/packages](https://github.com/dhis2/packages) was copied to the `packages/d2-repo` folder, with some modifications to make it work with the CLI system.  Once the CLI utils package is published we could alternatively just modify [dhis2/packages](https://github.com/dhis2/packages) in place and maybe rename it to `cli-packages` (or just leave it).  The alternatives would be to either archive that package or make it into a simple CLI entrypoint wrapper around the `d2 packages` command (see `packages/d2-repo/bin/d2-repo`), though that seem superfluous.

Would you prefer keeping the `dhis2/packages` name for that standalone tool and linking it in as a sub-command so it's also accessible as `d2 packages`?

README.md
amcgee: I think `d2 packages` (alias `pkgs`) works.

amcgee:
> We could take a hint from Babel's playbook and call our subpackage `helper-utils` to not squat the `d2-utils` namespace, which gives us a complete namespace to use for internal packages: `@dhis2/cli-helper-utils`, `@dhis2/cli-helper-something`. I think it's better to squat the `d2 helper` namespace than to squat the `d2 utils` namespace. We might want to use `d2-utils` but I'm not seeing `d2-helper` being necessary.

I like that.  I wasn't really happy with the `utils` name, actually.  It's technically the engine for d2 CLIs (individual CLIs interact with a wrapped `yargs` but *only* need to depend on the one `cli-utils`/`cli-helpers-engine` package.  Maybe `cli-helpers-engine`?

varl:
> Would you prefer keeping the `dhis2/packages` name for that standalone tool and linking it in as a sub-command so it's also accessible as `d2 packages`?

I don't mind either way, depends on if we want to contain everything in here for ease-of-development. I would prefer if we either _copy-and-archive_, so `dhis2/packages` is inline in the CLI-repo, or that we use `dhis2/packages` as a plugin and rename it to `cli-packages` so it's clear the entire thing is a sub-command.

varl:
> I was thinking about the app namespace as everything related to DHIS2 apps, not just the application repositories themselves. So, for instance, `d2`, `d2-ui`, `d2-i18n` etc etc all fall under the "app" umbrella. That could be more confusing than helpful, though. I was trying to limit top-level namespaces, at least initially.

Alright, that's fine I think, although having an alias for `lib` off the bat makes it easier to split them later and not change configuration except the `d2-cli` version. Maybe that's a too trivial issue to care about at this point.

> > `code-style`, or `commit-style` are technically separate but do the same.
> 
> Thoughts on where these might live if not under the packages namespace? We can keep them there for now.

I can easily see us extending the code style part to more than just JS so perhaps `d2-style` could work.

Usage would be:

* JS: `d2 style js`
* Git: `d2 style commit` (or `d2 style git commit` if we want to be explicit)
* CSS: `d2 style css`

Heck we might even establish a `d2 style bash` for [Shellharden](https://github.com/anordal/shellharden), and eventually we might even have a `d2 style java` for the java devs...?

varl:
> Maybe `cli-helpers-engine`?

This sounds reasonable. 👍

amcgee:
Switched to Lerna, updated namespaces to `packages` and `style`, implemented husky hooks to dogfood `d2 style js` and `d2 style commit`

`d2-packages` is currently in-lined here but we could easily make it a plugin and rename the existing repo/package if we want to keep the history there.

amcgee:
> > Maybe `cli-helpers-engine`?
> 
> This sounds reasonable. 👍

Done

amcgee:
@varl any more comments on this or should I merge it, publish the packages in the chosen namespaces, and iterate?

varl:
@amcgee merge it & iterate.

>d2-packages is currently in-lined here but we could easily make it a plugin and rename the existing repo/package if we want to keep the history there.

I can archive the dhis2/packages and dhis2/code-style repos after updating the README to direct readers this way. The history can remain there.

amcgee:
Removed d2-packages for now, will re-add once the helper engine is published and the packages repo is renamed.

PR #4, opened 2019-02-01
ci: setup travis

varl:


.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: node_js
+node_js:
+- 10
+before_script:
+- npm install --global @dhis2/deploy-build
+script:
+- echo "no build script, proceed..."
```
varl: Once we have some tests we should trigger them here, right now this just deploys the packages one-by-one.

The script reads the version from `package.json` of each package, but I think Lerna doesn't explicitly set versions in `package.json` when publishing?

PR #5, opened 2019-02-05
fix: add d2-style as external plugin

varl:
depends on https://github.com/dhis2/code-style/pull/10 being merged and published.


amcgee:
(see comments on dhis2/code-style#10)

PR #7, opened 2019-02-05
feat: add uid generation to d2-script

varl:
This ports the functionality from @dhis2/dhis2-cli to a script command module in @dhis2/cli.

It is backed by the @dhis2/dhis2-uid generation library.

**Question:**
Should we call it `d2-script` or `d2-scripts`?

Usage now is `d2 script uid`.


amcgee:
Nice!

@varl I think I prefer `d2-scripts` because it subtly implies that each script is discrete and self-contained, rather than something like `d2 app` which is a scope-narrowing namespace.  That's fighting against consistency, though, as other top-level namespaces are singular.  Does the distinction make sense to you or is it just confusing?

For the command name, I think we should try to stick to the convention `d2 <noun>.. <verb>`, so something like `genuid` or `generate-uid` or `uid generate` might work here.  I added a note to the README to do this for `d2 style` as well.

Another point to consider - we don't necessarily have to expose every namespace as its own standalone CLI.  Having `d2-scripts` and `d2-app` and `d2-style` etc. might be more confusing than helpful since they're all accessible through `d2` itself, so maybe we should funnel users through the `d2` cli instead?  In this case it might make sense if there's anyone using `@dhis2/dhis2-cli` for UID generation today, then we get an almost-dropin-replacement (though not quite).  Thoughts?

varl:
Yeah, I agree that it's unnecessary to create a new namespace for every little thing. I think we kicked around the idea of breaking down those silos a bit more, e.g. we could have a `generate` namespace which has command modules for `uid`, etc.

```
d2 generate uid
d2 generate app
```

varl:
After some discussion on Slack:

- Rename `d2-script` to `d2-utils`
- Use a noun-verb command structure

PR #12, opened 2019-03-01
Release module

varl:
This moves the `d2-packages release` command into `d2-app` in its current form.

It does not support Lerna enabled packages yet, but it's main purpose is for everything else.

https://github.com/dhis2/cli-packages/pull/13

### DIFF
--- a/packages/d2-app/index.js
+++ b/packages/d2-app/index.js
@@ -1,9 +1,11 @@
 const { namespace } = require('@dhis2/cli-helpers-engine')
 const createApp = require('@dhis2/create-app')
-const appScripts = require('./src/scripts')
+
+const appScripts = require('./src/scripts.js')
+const release = require('./src/release.js')
 
 module.exports = namespace('app', {
     desc: 'Manage DHIS2 applications',
     aliases: 'a',
-    commands: [createApp, appScripts],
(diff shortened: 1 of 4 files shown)


varl:
>1\.This might be better co-located with the (currently not integrated) other app scripts (`start` and `build` etc.) rather than `d2 app`?

Oh yeah, once we have the start/build app scripts integrated this should be moved there.

>3\. I think we should consider consolidating the `release`, `deploy`, and `publish` commands - they could do slightly different things depending on whether the current context is an app, lib, or monorepo but should probably be combined so you can't skip one or run them out of order accidentally.

`release` and `publish` are definitely tied to that order. We can revisit and see if we can avoid running `release` locally and `publish` remotely, but that's the reason they are separate today. 

`deploy` is a bit different and is not tied to the the release process, it's primary job is making sure that we have app artifacts that can be pulled in by the backend.

It deploys other libs to d2-ci as well, mostly to avoid duplicate code between `deploy-build.sh` and `publish-build.sh`, and also to keep the build pipeline as similar as possible.

Although I use the d2-ci builds to verify lib PRs (d2-ui, ci, etc.) locally by just pulling the ready-built artifact into apps to avoid switching branches locally and linking. I might be alone in that regard.

PR #13, opened 2019-03-01
Why lerna tho?

varl:
So I've been using this repo for a while now without Lerna and instead relying solely on Yarn Workspaces.

Publishing-wise, we are relying on [`publish-build`](https://github.com/dhis2/deploy-build/blob/master/publish-build.sh) which also doesn't need Lerna to do what we want.

We aren't using `lerna bootstrap` to link the packages, we aren't using `lerna publish` to publish the packages... 

If we remove Lerna we can also use the same `release` command for this package.

So why Lerna, tho? I don't see the use case for it.


varl:
I found the use case, we use it to update the sub-packages version strings, so we need the `deploy-build` step for that to be done automatically in this case: https://github.com/dhis2/deploy-build/blob/master/deploy-build.sh#L169

varl:
I think it gives us a nice workflow using just Yarn and eventually the `d2-app release` script for changelog generation.

amcgee:
Side note : we have only Branch builds enabled in Travis.com currently.  I think we should enable PR builds (I've set that as the required check) to support PRs from external forks, and I don't think we need both Branch and PR builds for local PRs (don't know if it's possible to de-duplicate).  We should figure out our preferred configuration across repos and be consistent.

varl:
>I'm not sure how I feel about pushing build artifacts to d2-ci (it's identical to the code currently, so redundant with this repo)

There's one difference: the `version` property in the sub-packages `package.json` is set by the `deploy-build` script. Not a big deal of course.

>I do think we should look into combining `release`, `deploy`, and `publish` scripts since they're all kind of variations on a theme and confusingly-named.

Yeah, we should look into restructuring those scripts when we port them to Node. I do like that they are distinct operations, but no need for them to be isolated scripts in different places.

varl:
>Side note : we have only Branch builds enabled in Travis.com currently.

I set it to branch builds because that builds each commit as-is, as the PR builds the merge between the branch and the base. Branch builds gives us a way to use the builds on d2-ci on a per commit basis. We could use both branch builds and pr builds in Travis if we modify the `deploy-build` script to ignore PR builds.

> I think we should enable PR builds (I've set that as the required check) to support PRs from external forks, and I don't think we need both Branch and PR builds for local PRs (don't know if it's possible to de-duplicate).

The external forks point is a good one.

varl:
Build failed because the `GITHUB_TOKEN` wasn't set up for Travis.

PR build skips deployment: https://travis-ci.com/dhis2/cli/builds/102778345
Branch build runs deployment: https://travis-ci.com/dhis2/cli/builds/102945248

varl:
I don't know why "Travis CI - Pull Request" is waiting for status to be reported even though the build succeeded?

amcgee:
@varl PR builds were disabled for this repo in Travis (don't know why), I've enabled them and restarted the PR build which hopefully will trigger an update of the status check?

amcgee:
Had to push an innocuous case change to the `No build script` message in `.travis.yml` but the builds (both PR and branch) were triggered and status updated once I toggled PR builds on at travis.com

PR #15, opened 2019-03-05
chore: fix greenkeeper.json

amcgee:
Use 4-space tabs and add missing package.json reference

### DIFF
--- a/greenkeeper.json
+++ b/greenkeeper.json
@@ -1,14 +1,15 @@
 {
-  "groups": {
-    "default": {
-      "packages": [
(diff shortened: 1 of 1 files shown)


varl:
I changed `d2 style js` to only hit javascript files btw, so when greenkeeper re-enabled on this repo it wrote it with it's own format (2-spaces).

amcgee:
@varl yep, `d2 style js check` only hits javascript files but `d2 style js apply` hits css and json as well (see https://github.com/dhis2/cli-style/issues/20)

PR #17, opened 2019-03-19
feat: revamp cluster management

amcgee:
This change revamps cluster management to allow full end-to-end cluster control.  To spin up a new cluster with the 2.31.1 DHIS2 core version, simply run:

```sh
> d2 cluster up dev --seed
> d2 cluster logs dev core
# Wait for the line "Catalina.start Server startup in XXXX ms", then ctrl+c / cmd+c to terminate
> d2 cluster restart dev gateway # this is a hack necessary to rehup the gateway 
# DHIS2 is available with Sierra Leone db at http://localhost:8080
# Run the Analytics Table export task from the Data Administration app
```

Note that any Docker image in [amcgee/dhis2-core](https://cloud.docker.com/u/amcgee/repository/docker/amcgee/dhis2-core) can be used, just omit the `-alpine` suffix.  Sierra Leone demo databases are automatically downloaded from the [dhis2-demo-db](https://github.com/dhis2/dhis2-demo-db) repository.

**Known issue** The 2.31 patch releases are nested, so they cannot be downloaded automatically.  You can download any .sql.gz file manually and specify it with the `--seedFile` option to the `d2 cluster up` command or the `--path` option to the `d2 cluster seed` command

**Known issue** The `d2 cluster seed <version>` command sometimes exits prematurely after downloading the db archive, run it again _without_ the `--update` option to continue seeding the cluster database from the downloaded file.

TODO
- [x] Automatically seed when running `up`
- [ ] (optional) Wait for tomcat to start, restart gateway server, and notify user of availability at URL
- [ ] (optional) use real verbs for commands `up`, `down`, `logs`?
- [ ] (optional) Pass cluster version before verb?
- [x] Document


amcgee:
@jenniferarnesen @varl @vilkg I requested reviews from you so that you're aware of these new `d2 cluster` features.  It could use a little more polish, but we're getting really close to a one-liner infrastructure setup for any version of DHIS2.

Comments and feedback welcome!

varl:
>(optional) Pass cluster version before verb?

Do you mean like this?: `d2 cluster <version> logs`

I much prefer the form in place now: `d2 cluster logs <version>`

The latter form is more ergonomic when you want repeat the same command over multiple versions without using shell replacement tricks (e.g. `^dev^2.31.1` to replace `dev` in the last command with `2.31.1`).

amcgee:
>     * One thing I would like is a `build` command where I could pass in a custom `apps-to-bundle.json` file through a parameter, and have the core use that file when it builds the Docker image, which it then would use to spin up and instance.

I've thought about that, but I think I would prefer this:
1. Support a CLI option to install a local development build of an app (or arbitrary tarball) into a running Core server, something like `d2 cluster apps install dev ./build`.  That would then use the app store API (or run a local dev version of the app store server) to add the app to the core instance.
2. Possibly distribute a zero-app Core image on Docker Hub to which you could then pass an `apps-to-bundle.json` and install a suite of applications after a much slimmer base install of the Core WAR file.

@varl Do you see custom core builds as important to support if we're going to move away from bundling apps?  FWIW you can do what you're asking for now as follows:

```
> cd dhis2-core
# edit apps-to-bundle.json in the core repo
> docker build . -t amcgee/dhis2-core:local-alpine
> d2 cluster up local
```

It's not super clean but it should do the job.  Note that we'll hopefully move the docker image repo to a dhis2 namespace soon.

amcgee:
> > (optional) Pass cluster version before verb?
> 
> Do you mean like this?: `d2 cluster <version> logs`
> 
> I much prefer the form in place now: `d2 cluster logs <version>`
> 
> The latter form is more ergonomic when you want repeat the same command over multiple versions without using shell replacement tricks (e.g. `^dev^2.31.1` to replace `dev` in the last command with `2.31.1`).

Yeah, that's what I meant.  I think it bugs me because it breaks the law of increasing contextual specificity (you're operating on a particular named cluster, not the cluster namespace itself) - but at the same time injecting effectively dynamic namespacing could be problematic too.

It gets a little unwieldy when you have to pass other positional arguments to a command, for instance:

```
d2 cluster restart <dev> [gateway] # options in brackets
d2 cluster apps add <dev> [maps] [dashboards] ...
```

In that last one, the `d2 cluster dev d2 apps add` makes a bit more sense to me.  I could see us having:

```sh
d2 cluster list
d2 cluster add dev
d2 cluster dev up|down|seed|restart [service]|logs [service]|...
d2 cluster add 2.30
d2 cluster 2.30 up|down|seed|restart [service]|logs [service]|...
```

I see your point though, don't think it's a huge loss to skip that feature.

varl:
> I've thought about that, but I think I would prefer this:

OK, yeah, either of those would work for me.

> @varl Do you see custom core builds as important to support if we're going to move away from bundling apps?

Yes. 

- It is not decided that we are going to split the backend and frontend retroactively for the supported 4 versions (e.g. right now would be 2.31, 2.30, 2.29, 2.28) so we are going to need this mechanism for a few years.
- We will still want to be able to produce a "fat WAR" for those situations where DHIS2 is deployed offline.
- To start, we will still have to bundle a bare minimum amount of apps that are required to set up DHIS2 (maintenance app, users app, etc.) even if we move the bulk to the app store.

> 
> ```
> > cd dhis2-core
> # edit apps-to-bundle.json in the core repo
> > docker build . -t amcgee/dhis2-core:local-alpine
> > d2 cluster up local
> ```
> 
> It's not super clean but it should do the job. Note that we'll hopefully move the docker image repo to a dhis2 namespace soon.

This is perfectly fine! No need for the feature then, this is workable.

amcgee:
> Yes.
> 
>     * It is not decided that we are going to split the backend and frontend retroactively for the supported 4 versions (e.g. right now would be 2.31, 2.30, 2.29, 2.28) so we are going to need this mechanism for a few years.
> 
>     * We will still want to be able to produce a "fat WAR" for those situations where DHIS2 is deployed offline.
> 
>     * To start, we will still have to bundle a bare minimum amount of apps that are required to set up DHIS2 (maintenance app, users app, etc.) even if we move the bulk to the app store.

Cool, makes sense

varl:
> I see your point though, don't think it's a huge loss to skip that feature.

Yeah, I see the problem. I didn't think about commands as complex as that. :thinking:

amcgee:
@Philip-Larsen-Donnelly might be interested in this as well

Philip-Larsen-Donnelly:
Yes, he is.

amcgee:
@Philip-Larsen-Donnelly welcome!  Any thoughts before making this available?  Or just glad to know it's a thing?  FYI we're planning to cut a 1.0 release of the CLI soon with this feature included

Philip-Larsen-Donnelly:
Mostly glad to know it's a thing. 

I've been working on a set of ansible playbooks and roles for managing our infrastructure. "Containerized" deployment of instances should eventually be an option (and may even be the default), alongside "native" deployment, so I am keen to stay in the loop.

Managing the play and qa infrastructure in this way highlights certain challenges, which we may, or may not, wish to address with the "containerized" versions too. e.g.:

- zero-downtime war updates: (I've already tested a native approach with tomcat that I plan to use as standard on the new "play").
- centralised logging: as a baseline, I have set up an ELK server and am defining appropriately-configured "beats" on the deployed instances to feed the central log server.
- ...and so on.

But none of these things need to be addressed before the CLI is more widely available.

amcgee:
Very cool!

From the containerized perspective:
1. Zero-downtime updates should be possible as-is by spinning up a new
container and updating the pointer in nginx.  It could be streamlined a bit
though
2. I've thought about adding the ability to mix and match different cluster
configurations, including one which adds ELK stack log aggregation and
analytics.  It would be pretty straightforward to capture logs from all the
running containers.
3. I'd like to support auto-scaling and service discovery on docker stack
or kubernetes, which shouldn't be too hard but I haven't looked into too
far yet.

Related to this PR, is there any reason we can't move
https://github.com/dhis2/dhis2-demo-db/tree/master/sierra-leone/2.31/2.31.0
up one level so it can be auto-discovered?  Also, could we have a 1-to-1
mapping of core versions to demo database versions even if there are no db
updates?  (there's no 2.31.1 for instance)

Philip-Larsen-Donnelly:
One of the things I like about the zero-downtime updates with tomcat is that you basically run in parallel; Browser sessions that are using the old war version continue to use that version, while new sessions will be directed to the new war version. When there are no sessions using the old version any longer, tomcat will (if configured) remove the old war version. Maybe something similar can be achieved with some nginx magic too!?

Let's have a chat so that I can understand the requirements for auto-discovery and we can discuss possible structures.

amcgee:
I'm going to merge these changes as-is, opened #21 for follow-on enhancements.

@Philip-Larsen-Donnelly that's a bit farther off but sure happy to chat

PR #19, opened 2019-03-21
feat: add semantic-release to d2-app scripts

varl:
This adds the `release` command under `d2-app scripts`. It is based on `semantic-release` so it is fully automated. Right now it doesn't allow overrides to its configuration from the application/library and uses "sane defaults" (i.e. the way we want it to behave). When it's proven to work we can add the ability to merge configurations from the application/library.


packages/d2-app/src/commands/scripts/release.js
```diff
@@ -0,0 +1,82 @@
...
+        ],
+
+        '@semantic-release/npm',
+
+        [
+            '@semantic-release/git',
+            {
+                assets: ['CHANGELOG.md', 'package.json', 'yarn.lock'],
+                message:
+                    'chore(release): cut ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}',
```
varl: https://changelog.travis-ci.com/skip-travis-ci-builds-with-skip-travis-in-commit-message-74620

Not sure if we want this to skip CI. :thinking:

packages/d2-app/package.json
```diff
@@ -1,18 +1,22 @@
...
-    "description": "D2 CLI module for interacting with frontend applications",
-    "main": "index.js",
-    "repository": "https://github.com/dhis2/cli",
-    "author": "Austin McGee <austin@dhis2.org>",
+    "bin": {
+        "d2-app": "./bin/d2-app"
+    },
+    "version": "1.0.0",
+    "main": "src/index.js",
+    "author": "DHIS2 &lt;info@dhis2.org&gt;",
```
amcgee: This is a bug in the `create` template, need to un-escape it somehow

amcgee: But for now maybe manually is fine?

packages/d2-app/src/commands/scripts/release.js
```diff
@@ -0,0 +1,82 @@
...
+            ...process.env,
+            GIT_AUTHOR_NAME: '@dhis2-bot',
+            GIT_AUTHOR_EMAIL: 'ci@dhis2.org',
+            GIT_COMMITTER_NAME: '@dhis2-bot',
+            GIT_COMMITTER_EMAIL: 'ci@dhis2.org',
+        },
+    }
+
+    try {
+        const result = await semanticRelease(options, config)
```
amcgee: We should add a `--dry-run` CLI option which passes through to the same option for semantic release, so we can test release logic locally

packages/d2-app/src/commands/scripts/release.js
amcgee: Hmm yeah probably fine for now?

amcgee:
Really cool to get this all automated!

packages/d2-app/src/commands/scripts/release.js
varl: I can do that. Right now you are barred from doing a local release as it checks for a CI environment by default. I'll make that explicit too.

amcgee:
Also also - how do we do this in mono-repos?  Would love to use it in `dhis2/cli` itself, but I think we'd need to restructure the repo, use [this](https://github.com/Updater/semantic-release-monorepo) which doesn't look super solid, or make our own set of configs for mono-repo support (and particularly single-version monorepo support?)

packages/d2-app/package.json
varl: 86c2edd

packages/d2-app/src/commands/scripts/release.js
varl: Introduced `--no-dry-run`/`--dry-run` and `--no-ci`/`--ci`

amcgee: Do you think we could get away with not allowing `--ci`?  Maybe we need it

varl:
> Also also - how do we do this in mono-repos? Would love to use it in `dhis2/cli` itself and with `d2-ui` for instance, but I think we'd need to restructure the repo, use [this](https://github.com/Updater/semantic-release-monorepo) which doesn't look super solid, or make our own set of configs for mono-repo support (and particularly single-version monorepo support?)

Yeah, I thought I read about this working, but I was mistaken: https://github.com/semantic-release/semantic-release/issues/193

This is likely more solid, but it uses commitizen: https://github.com/atlassian/lerna-semantic-release

packages/d2-app/src/commands/scripts/release.js
varl: We can get away with not having `--dry-run` and `--ci`, I think.

`--ci` defaults to `true` in semantic-release, so the default is to only allow CI environments to run the release.

So if I run the release command just the defaults it automatically goes into dry run mode.

```
d2 app scripts release
...
[1:59:08 PM] [semantic-release] › ⚠  This run was not triggered in a known CI environment, running in dry-run mode.
```

To actually run the `release` locally I need to pass in `--no-ci`.

amcgee:
Oh cool!  I thought we had to pass --dry-run but if it skips when not in a
CI that's even better.  Sorry for the misdirection

packages/d2-app/src/commands/scripts/release.js
varl: >Also, is this actually for apps or just for libraries? Apps still have a different (non-npm) release process right?

Yeah, I'm pondering adding a `--mode [app/lib]` switch, as the only thing we would need to _not_ do for apps is publish to NPM:  https://github.com/dhis2/cli/blob/86c2edd83d92d72f08115955c0e0b377c2591d4c/packages/d2-app/src/commands/scripts/release.js#L24

amcgee:
What about a `--publish` flag?

packages/d2-app/src/commands/scripts/release.js
amcgee: What about a `--publish` flag, set for libs but not apps?  Or check the `publishConfig` in package.json?

( GitHub hates email )

amcgee: > So if I run the release command with the defaults it automatically goes into dry run mode.

Cool, yeah no need for dry-run then!  Sorry I didn't realize you could preview it locally without

varl: Sure. :+1:

An alternative is just setting `"private": true"` in `package.json` for Apps (not verified).

varl: I would like to have a `--publish [target]` in that case, so we can publish to NPM for libs and App Store for apps.

varl: 0c6fd99

varl:
Review notes:

- dryRun by default on local machines
- Do not publish by default
- Allow for publish providers through `--publish [provider]`
- Only provider now is NPM

packages/d2-app/package.json
```diff
@@ -1,18 +1,22 @@
...
+    "bin": {
+        "d2-app": "./bin/d2-app"
+    },
     "version": "0.11.0",
-    "description": "D2 CLI module for interacting with frontend applications",
-    "main": "index.js",
-    "repository": "https://github.com/dhis2/cli",
-    "author": "Austin McGee <austin@dhis2.org>",
+    "main": "src/index.js",
+    "author": "Viktor Varland <viktor@dhis2.org>",
```
amcgee: I see what you did there ;-)

varl: Hahaha, oh man, sorry about that.

I did a `rm` of `d2-app`, then regenerated it with the `create cli` tool to allow it to match the new convention as it seemed to predate the create script, and accepted the defaults. :D

packages/d2/src/index.js
```diff
@@ -4,11 +4,12 @@ const command = namespace('d2', {
     desc: 'DHIS2 CLI',
     builder: yargs => {
         yargs.commandDir('commands')
-        yargs.command(require('@dhis2/cli-create'))
-        yargs.command(require('@dhis2/cli-app'))
         yargs.command(require('@dhis2/cli-cluster'))
+        yargs.command(require('@dhis2/cli-packages'))
```
amcgee: Get outa here!

packages/d2-app/src/commands/scripts/release.js
amcgee: This solution looks good to me!

packages/d2-app/package.json
varl: 8352ec4

packages/d2/src/index.js
varl: What the!

e4e4f53

PR #24, opened 2019-03-23
feat: semantic release update deps

amcgee:
So... I may have gone a little overboard but I think this gets us all the way there. 

This looks through glob patterns in the `workspaces` property of the root `package.json`.  It then loops through all the package.jsons it can find and updates local dependencies (other packages found in the workspace) to `nextRelease.version` (using a `^` range).  It also updates the `version` property in each package.json, including the root.

~Right now this will try to publish the root package as well, which should fail because it's private.  We might want to filter out private repos before trying to publish them (or at least omit the root one in this specific case).~ = **done**

Also - I don't think we're actually updating the `yarn.lock` files anywhere, but this should find and push them all in the `git` stage if they've changed

Now time for 🛌

### DIFF
--- a/packages/d2-app/src/commands/scripts/release.js
+++ b/packages/d2-app/src/commands/scripts/release.js
@@ -1,23 +1,17 @@
 const { reporter } = require('@dhis2/cli-helpers-engine')
-
-const { readdirSync } = require('fs')
-const { join } = require('path')
-
+const { existsSync } = require('fs')
+const path = require('path')
 const semanticRelease = require('semantic-release')
+const getWorkspacePackages = require('./support/getWorkspacePackages')
 
-function publisher(target = '') {
+function publisher(target = '', packages) {
     switch (target.toLowerCase()) {
         case 'npm': {
-            return ['@semantic-release/npm']
-        }
-
-        case 'mono-npm': {
-            const packages = readdirSync('./packages')
-            return packages.map(p => {
+            return packages.map(pkgRoot => {
                 return [
                     '@semantic-release/npm',
                     {
-                        pkgRoot: join('./packages', p),
+                        pkgRoot,
                     },
                 ]
             })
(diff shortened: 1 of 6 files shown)


packages/d2-app/src/commands/scripts/release.js
```diff
@@ -29,38 +32,69 @@ function publisher(target = '') {
...
+        ...publisher(publish, packages),
+        gitPlugin,
+        '@semantic-release/github',
+    ]
 
     const options = {
-        branch: 'master',
-        version: 'v${version}',
-        plugins: plugins.filter(n => n),
+        branch: 'feat/semantic-release-update-deps',
```
varl: Danger zone!

packages/d2-app/src/commands/scripts/release.js
```diff
@@ -29,38 +32,69 @@ function publisher(target = '') {
...
+        version: 'v${version}-test',
```
varl: This too

packages/d2-app/src/commands/scripts/support/semantic-release-update-deps.js
```diff
@@ -0,0 +1,93 @@
+const fs = require('fs')
+const path = require('path')
+const SemanticReleaseError = require('@semantic-release/error')
+const AggregateError = require('aggregate-error')
+const normalizeAndValidatePackages = require('./normalizeAndValidatePackages')
+
+const verifyConditions = (config = {}, context) => {
+    const { silent, packages } = config
+    const { logger } = context
+    if (!packages || !packages.length || packages.lenght < 2) {
```
varl: `packages.lenght` should be `packages.length`

packages/d2-app/src/commands/scripts/release.js
```diff
@@ -29,38 +32,69 @@ function publisher(target = '') {
...
+                    ),
+            ],
             message:
                 'chore(release): cut ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}',
         },
-    ])
+    ]
 
-    plugins.push('@semantic-release/github')
+    // Order matters here!
```
varl: :+1:

packages/d2-app/src/commands/scripts/support/semantic-release-update-deps.js
amcgee: Oof that's embarrassing.  Typescript would have caught this, but I'll settle for eagle-eyed @varl for now.

amcgee:
@varl thanks for the review, nice catches!  Fixed them now.

PR #25, opened 2019-03-25
fix: don't update the package.json version before npm stage

amcgee:
Pass `NPM_CONFIG_ALLOW_SAME_VERSION` to `npm publish` so it doesn't fail when we've already updated the package.json version in the updateDeps plugin.


amcgee:
🐦

PR #37, opened 2019-05-06
chore: pin all @dhis2 scoped package versions

varl:
Change to pinned versions for `@dhis2`-scoped packages to allow Greenkeeper to keep them updated for us.

Also update `@dhis2/cli-style` to get the new lint rules pulled in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     ],
     "license": "BSD-3-Clause",
     "devDependencies": {
-        "@dhis2/cli-style": "^3.0.0",
+        "@dhis2/cli-style": "3.1.1",
         "husky": "^1.3.1"
     },
     "husky": {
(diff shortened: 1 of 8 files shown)


amcgee:
@Mohammer5 agreed, that probably makes this a minor bump instead of a "chore"?

Mohammer5:
> @Mohammer5 agreed, that probably makes this a minor bump instead of a "chore"?

@amcgee I don't know, not necessarily. Commit message can stay `chore` and if nothing changes functionality-wise, I wouldn't say it's worth to increment the minor version number either. 
On the other side, better safe than sorry?

varl:
> @varl it looks like you're not working off the latest master branch, the current version of CLI packages in this repo are all `v1.0.2`

Whoops!

varl:
I think the pin change is a `chore` (doesn't need a release).

Changing the version of `cli-style` warrants a release thought because it's actually a `feat` because of the change in `cli-style` which adds functionality for ESLint, etc.

varl:
Closing this, and re-opening a new PR.

PR #38, opened 2019-05-06
chore: pin @dhis2-scoped package versions

varl:
Replaces: #37

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     ],
     "license": "BSD-3-Clause",
     "devDependencies": {
-        "@dhis2/cli-style": "^3.0.1",
+        "@dhis2/cli-style": "3.0.1",
         "husky": "^2.0.0"
     },
     "husky": {
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -12,7 +12,7 @@
     "license": "BSD-3-Clause",
     "private": false,
     "dependencies": {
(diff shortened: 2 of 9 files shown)


varl:
This only pins versions, and we can let Greenkeeper update to see that everything works as intended.

amcgee:
@varl nice.  Probably good to also update the reference to `@dhis2/cli-style` in the root `package.json` (used by the pre-commit hook) so that we don't have both version ranges in the yarn.lock file.  There's still a carat-version cli-helpers-engine which I believe comes from `cli-style`, so  hopefully that will drop out when greenkeeper updates `cli-style`

varl:
Search and replace fail on this branch. Fixeed in c88b526.

amcgee:
The `yarn.lock` update seems excessive (dozens of added deps) and still include the `cli-style@^3.0.2` entry...?

varl:
From what I can trace, most of the added deps are from `semantic-release@^15.13.3`, which is also added by this PR. Which I do not understand. We should have that already.

varl:
Running `yarn install` on master updates the `yarn.lock` as well:

[yarn.lock.txt](https://github.com/dhis2/cli/files/3153303/yarn.lock.txt)

amcgee:
This looks OK, but we'll actually need to change the release script as well since it automatically updates dependency versions using ^ ranges in a monorepo

varl:
Done in d00acd7.

PR #39, opened 2019-05-09
fix: expose the version as an environment variable in the context

varl:
~Relates to: https://github.com/amcgee/dhis2-backend/pull/2~

No longer coupled to the dhis2-backend PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.0.2](https://github.com/dhis2/cli/compare/v1.0.1...v1.0.2) (2019-03-28)
+
+
+### Bug Fixes
+
+* upgrade @dhis2/cli-helpers-engine and @dhis2/cli-style ([#35](https://github.com/dhis2/cli/issues/35)) ([251ac22](https://github.com/dhis2/cli/commit/251ac22))
+
 ## [1.0.1](https://github.com/dhis2/cli/compare/v1.0.0...v1.0.1) (2019-03-27)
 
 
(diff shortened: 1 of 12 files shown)


amcgee:
Very cool that this enables support for multiple instances behind a gateway!  See [my comment on the companion pr](https://github.com/amcgee/dhis2-backend/pull/2#pullrequestreview-235472970) for concerns about tight coupling.

One question which may be important - do we want to support arbitrary context paths per cluster, or would it be sufficient to support a singular "cluster" gateway which proxies to multiple "versioned and named" clusters behind it.  I.E.

```
d2 cluster config --port 8080
d2 cluster add --version dev --name mydevinstance
d2 cluster add --version 2.32 --name latest
```
(or something better with nicer commands)

Could result in a single gateway pointing at multiple clusters behind-the-scenes, so you could access `localhost:8080/mydevinstance` and `localhost:8080/latest`.  This is a little less flexible but a little more controlled than allowing arbitrary context specification.

varl:
Could you take another look at this as well as https://github.com/amcgee/dhis2-backend/pull/2#issuecomment-490872497? They should be less coupled.

varl:
>One question which may be important - do we want to support arbitrary context paths per cluster, or would it be sufficient to support a singular "cluster" gateway which proxies to multiple "versioned and named" clusters behind it. I.E.

I'm not exactly sure how this would work in practice, for example, I would still have to put an nginx reverse proxy in front of that "singular gateway" in order to serve the other sites I have on the same server.

But maybe `--name` is better than `--context`, as a different `docker-compose` repo might want to use the name environment for something else.

varl:
I pulled out the single port discussion to a separate issue.

varl:
This only exposes the version as an environment property now.

PR #43, opened 2019-05-10
feat: schema differ

Birkbjo:
Adds 'schema' namespace with command 'diff' under utils. I don't know if we want this in its own repo, or directly in utils like I did here. Would not be hard either way to extrat to its own repo or package.

The background for making this is that we often get weird bugs (especially in the maintenance app) when seemingly small changes to properties in the schemas (like setting ie embedded: false) are made. They are hard and time consuming to debug, and we frontend developers always get the blame for broken apps, even if the cause is an undocumented API change.

The idea is that we can run this tool to grab and generate a diff of the schemas. It outputs an html file, with all the schemas and delta-data embedded (much like source-map-explorer), so its easily sharable as a static resource.

I'm not sure if this is going to be run in the build pipeline. I initially created a prototype to debug a weird bug (https://jira.dhis2.org/browse/DHIS2-6767) , and then decided to port it here, as I believe its a valuable utility. 

I tried to minimize typing long urls, so I made the play-server the default base-url, and that will be used if the --base-url does not have an argument.
 So you can do this: `d2 utils schema diff /2.31dev /2.32 -b`.

Example diff: https://birk.dev/static/diffs/2.32.0_a9f597b__2.33-SNAPSHOT_9d8bb91.html


packages/create-app/bin/d2-create-app
```diff
@@ -5,4 +5,4 @@ const pkgJson = require('../package.json')
 const command = require(`..`)
 
 notifyOfUpdates(pkgJson)
-makeEntryPoint(command).parse()
+makeEntryPoint(command)
```
Birkbjo: Do we want to fix these in its own PR? This PR wont work without this fix though.

packages/utils/src/cmds/schema/diff/index.js
```diff
@@ -0,0 +1,269 @@
...
+    fs.writeFileSync(loc, html)
+    reporter.info('Visual output generated:', loc)
+    reporter.pipe(loc)
+}
+
+async function run({ url1, url2, baseUrl, html, json, ...rest }) {
+    cache = rest.getCache()
+    const left = await getSchemas(url1, baseUrl, rest)
+    const right = await getSchemas(url2, baseUrl, rest)
+    //let [left, right] = await Promise.all([prom1, prom2])
```
Birkbjo: I wanted to do this to send them concurrently. However, in the case of --prompt-auth we need to wait for the prompt for each request. Opted for simplicity instead of extra logic to fix this.

amcgee:
This looks awesome @Birkbjo, really useful!!  I'll take a closer look before Monday :-)

packages/create-app/bin/d2-create-app
amcgee: This is fine, thanks for finding the bug!

packages/utils/src/cmds/schema/diff/index.js
amcgee: Seems OK for now, might be nice to parallelize in the future

varl:
I expect that the default output of the command would be rendered straight to the terminal in a diff format similar to the `diff` command; unless that tool can be used, in which case that is preferable to me.

I don't like when tools write to disk automatically either, so I'd prefer to redirect `>` to a file if I need the output saved.

So my preferred usage (based on what I expect from a terminal application) would be:

```bash
d2 utils schema diff /2.31 /2.32
+ fieldName
- fieldName

d2 utils schema diff /2.31 /2.32 > result.diff

# render using `diff`
d2 utils schema diff /2.31 /2.32 | diff

d2 utils schema diff /2.31 /2.32 --json
{
    ...
}

d2 utils schema diff /2.31 /2.32 --json > result.json

d2 utils schema diff /2.31 /2.32 --html
<html>
...
</html>

d2 utils schema diff /2.31 /2.32 --html > result.html

# to open in default browser (linux)
d2 utils schema diff /2.31 /2.32 --html > result.html | xdg-open
```

amcgee:
I think I agree @varl about having CLI-first options, though I don't mind opening a browser with a `--open` flag, for instance.  I think I prefer `--format html` to `--html` but not super strongly.

varl:
`--format` could be nice just to minimize the amount of switches.

I'm OK with `--open` existing as a switch, I can still pipe it to different browsers if I need to through the saved `.html` file.

amcgee:
https://github.com/sindresorhus/open might be useful

Birkbjo:
Did a refactor now. --format instead of --html and --json. Now you can 
`d2 utils schema diff /2.31 /2.32 -b --format html > result.html`

By default it outputs to console, however I also added --output and -o to be able to generate the 'identifier'-name automatically. This outputs the filename, so you can do `d2 utils schema diff /2.31 /2.32 -b --format html -o | xargs open`

Some thoughts about the rendering: 
Previously all the data is outputted to html, and the formatting happens client side (by injecting the jsondiffpatch dep into the html).
Pros: 
- You can inspect the data in the console
- You have access to jsondiffpatch in the console, so you can format and patch as you wish
Cons:
- Bigger file size (we inject data and jsondiffpatch in the HTML)
- Minor loading time while the browser formats

I refactored this to generate and output just the diff to html. This means a reduction of ~90% of the outputted html size. From ~5MB to ~400kb. This means we now lose the original data, so you cannot see the data that was not changed, but I believe the reduction in size is well worth it.

Birkbjo:
> https://github.com/sindresorhus/open might be useful

I've looked into it. But I don't really want to add more dependencies than necessary, so I believe piping to `xargs open` or what other program your distro has is sufficient.

packages/utils/src/cmds/schema/diff/index.js
```diff
@@ -0,0 +1,314 @@
...
+    baseUrl: 'https://play.dhis2.org',
+    headers: {
+        'x-requested-with': 'XMLHttpRequest',
+        Authorization: utils.basicAuthHeader('admin', 'district'),
+    },
+    json: true,
+}
+
+let cache
+const prompt = inquirer.createPromptModule({ output: process.stderr })
```
varl: A prompt module seems like it belongs in `cli-helpers-engine` so we can be consistent in how we ask for user input.

packages/utils/src/cmds/schema/diff/utils.js
```diff
@@ -0,0 +1,30 @@
...
+    return Buffer.from(str).toString('base64')
+}
+
+function basicAuthHeader(user, pass) {
+    const b64 = btoa(`${user}:${pass}`)
+    return `Basic ${b64}`
+}
+
+function isAbsoluteUrl(url) {
+    return url.indexOf('http://') === 0 || url.indexOf('https://') === 0
```
varl: [`String.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) is a bit more contemporary than using `indexOf`.

packages/utils/src/cmds/schema/diff/index.js
varl: That's for a different PR anyway.

packages/utils/src/cmds/schema/diff/index.js
```diff
@@ -0,0 +1,314 @@
...
+    if (auth) {
+        ;[username, password] = auth.split(':')
+    }
+    if (auth === '') {
+        ;({ username, password } = await prompt([
+            {
+                type: 'input',
+                name: 'username',
+                message: `Username for ${url}`,
+                default: 'admin',
```
amcgee: This is fine for now but at some point we'll probably want to "centralize" the defaults for cli tools (where to find the server you want to talk to, what auth to use), so we can easily change/remove them or at least make them configurable.

Birkbjo:
Added support for fetching schemas, and saving these to disk. These files can be used as an input for the differ.
Sorry for all the refactoring... Had some common code I wanted to share. I was abit unsure of the folder structure, so don't hesitate to state your opinions about that. Maybe the the common schema code be moved to 'support' folder as well?

packages/utils/src/cmds/schema/diff/index.ejs
```diff
@@ -0,0 +1,57 @@
...
+            .left {
+                background-color: #fbb
+            }
+            .right {
+                background-color: #bfb
+            }
+        </style>
+         <%
+            const metaPropsToShow = ['contextPath', 'version', 'revision', 'buildTime']
+            const commitBaseUrl = 'https://github.com/dhis2/dhis2-core/commit/'
```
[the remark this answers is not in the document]
Birkbjo: It's from the /info endpoint. I'm unsure how it grabs it, as it's from the backend.

packages/utils/src/cmds/schema.js
```diff
@@ -0,0 +1,20 @@
...
+    builder: yargs =>
+        yargs
+            .option('force', {
+                type: 'boolean',
+                describe:
+                    'By default each schema is cached, identified by the version and revision. Use this flag to disable caching and download the schemas.',
+            })
+            .option('auth', {
+                type: 'string',
+                describe: `Auth to use, formatted as "user:password". Note that this is not safe, as password is shown in history.
```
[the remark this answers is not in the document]
varl: I think this would be a nice simplification.

varl:
I think that default should be to play, and the usage of the `--base-url`/`-b` would be to specify the base URL, e.g.:

`d2 utils schema diff /2.31 /2.32` hits play.dhis2.org.

`d2 utils schema diff /2.31 /2.32 -b dhis2.vardevs.se` hits my server.

`d2 utils schema diff dhis2.vardevs.se/2.31 play.dhis2.org/2.32` diffs the two different servers.

packages/utils/src/cmds/schema/diff/index.ejs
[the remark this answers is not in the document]
amcgee: @Birkbjo yeah, the problem is that when the image is built outside a git repo (including in our official docker build process) you get this from that endpoint:

```
  "revision": "fatal: Not a git repository (or any of the parent directories): .git",
```

That may be something we want to fix in the backend build process, but for now it's a problem if the backend is built in Docker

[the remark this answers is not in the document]
amcgee: ^ @varl we discussed this an aeon ago (my suggestion I believe was to pass the "build context" into the build rather than presuming anything about the environment in which it's built), might be worth revisiting now that we're adding Docker builds to our Jenkins pipeline?  cc @vilkg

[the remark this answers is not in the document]
varl: I remember. But what we discussed the same thing in a different context: https://github.com/dhis2/dhis2-core/pull/2607/files

The problem back then was when the scripts that bundle the application failed to put out the core's Git revision the build crashed. dhis2/core#2607 solved that by just setting the core revision to "n/a" if it fails.

This is the core pulling in revisions from the build environment: https://github.com/dhis2/dhis2-core/blob/667efb56e42659569e5904d92c492f57e6a4c7ea/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/system/DefaultSystemService.java#L206

Which is set in the main `pom.xml`: https://github.com/dhis2/dhis2-core/blob/0460082f65826af04d1298d97cc7bca2ef3f47af/dhis-2/pom.xml#L277-L283

That said: would be nice to get this "git agnostic" in both places.

Final note: to get fully git agnostic we would need to change the way the apps are downloaded from `git` to `http`.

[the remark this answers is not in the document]
amcgee: Yeah it was a slightly different context, you're right.  I agree it would be nice to move to a "stand-alone" build which only relies on the source code in the core repo.

In the meantime, can we make sure the schema diff doesn't break when `revision` is a git error message?

varl:
Use cases for this keeps popping up.. 👍

packages/utils/src/cmds/schema/diff/index.ejs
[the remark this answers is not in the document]
varl: This PR solves it so that the build stage 1 can access the Git repo and set the revision properly: https://github.com/dhis2/dhis2-core/pull/3377

amcgee:
This is looking really great @Birkbjo !!  I tested and it's working well, really love the HTML formatting of the diff.  The following are the final minor things, then we should be able to get this merged

One question: why are there all these "array reorder" diffs in what should be "identical" (or nearly identical) schemas?  Is array ordering non-deterministic in the backend or are these legit scheme divergences.

Screenshot 2019-05-28 11 35 38

And a couple future considerations for the CLI interface, probably not blocking this merge:
1. `--base-url` and `--auth` both apply to BOTH servers, which might be annoying when diffing i.e. play with a local server.  This can be worked around effectively by downloading schemas independently with `d2 utils schema fetch` and then diffing the downloaded files, so I don't think this is urgent.
2. I did find it somewhat cumbersome to "discover" the mechanism for generating and viewing the HTML diff, so it might be worth some documentation somewhere?

All in all really awesome tool, I think it will be incredibly useful.  Great work!

packages/utils/src/cmds/schema/index.js
```diff
@@ -0,0 +1,132 @@
...
+
+    const meta = await asyncRequest(infoUrl, requestOpts)
+    const rev = utils.isSHA(meta.revision)
+    if (!rev) {
+        meta.revision = 'REV-NA'
+        reporter.warn('No revision found')
+    }
+    const schemaFileName = `${schemaIdentifier(meta)}.json`
+    const loc = await cache.get(schemasUrl, schemaFileName, {
+        raw: true,
```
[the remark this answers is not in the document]
Birkbjo: Thanks for letting me know!

I still have a few finishing touches to do. Have been busy with some Maintenance-issues lately.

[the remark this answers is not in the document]
amcgee: Ok no worries, just ping me when it's ready!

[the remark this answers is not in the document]
Birkbjo: @amcgee , I think this is coming together now! Would appreciate a new peek, especially the docs, as things should hopefully be a bit clearer now.

amcgee:
@Birkbjo cool, I'll take a look at this in the next couple days!

packages/utils/src/cmds/schema/index.js
```diff
@@ -0,0 +1,143 @@
...
+                    )
+                process.exit(1)
+            }
+            return resolve(body)
+        })
+    })
+}
+
+function resolveConfig(args) {
+    if (!args.utils && !args.utils.schema) {
```
amcgee: This should be `||`, not `&&` - I get this error

```
TypeError: Cannot read property 'schema' of undefined
    at resolveConfig (/Users/awm/dev/dhis2/cli/d2/packages/utils/src/cmds/schema/index.js:47:36)
```

packages/utils/src/cmds/schema/index.js
```diff
@@ -0,0 +1,143 @@
...
+    const auth = {
+        username: serverConfig.username || conf.username,
+        password: serverConfig.password || conf.password,
+    }
+    if (auth.username && auth.password) return auth
+    return true
+}
+
+async function getAuthHeader(url, { auth }) {
+    let { username, password } = auth
```
amcgee: This breaks if the tool is used without a `config` file, which we should support.  After changing `&&` to `||` above I get this error:

```
./packages/utils/bin/d2-utils schema --auth diff https://play.dhis2.org/dev http://localhost:8080
d2-utils schema diff <leftUrl> <rightUrl>

Generate a diff of schemas between DHIS2 versions

Positionals:
  leftUrl   the url to the running DHIS2 server, should have schemas available
            relative to this at /api/schemas.json
            Can also be a JSON-file with schemas. See schema fetch
                                                             [string] [required]
  rightUrl  the url to another running DHIS2 server, should have schemas
            available relative to this at /api/schemas.json
            Can also be a JSON-file with schemas. See "utils schema fetch"
                                                             [string] [required]

Options:
  --version             Show version number                            [boolean]
  --config              Path to JSON config file
  --force               By default each schema is cached, identified by the
                        version and revision. Use this flag to disable caching
                        and download the schemas.                      [boolean]
  --auth                Prompt credentials for each server.
                                                      [boolean] [default: false]
  --base-url, -b        BaseUrl to use for downloading schemas. If this is set
                        leftServer and rightServer should be relative to this
                        url, eg. /dev.                                  [string]
  --output, -o          Specify the location of the output. If used as a flag a
                        file relative to current working directory is generated
                        with the name
                        "LEFT-version_revision__RIGHT-version_revision.html".
                        If the location is a directory, the default filename is
                        used and output to location.                    [string]
  --format              Specify the format of the output
              [string] [choices: "html", "json", "console"] [default: "console"]
  --ignore-array-order  Sort all nested arrays in schemas. The order of nested
                        arrays are non-deterministic, which may clutter the
                        diff. Enabling this will prevent most internal array
                        moves, which are probably irrelevant anyway.
                                                      [boolean] [default: false]
  -h, --help            Show help                                      [boolean]

TypeError: Cannot destructure property `username` of 'undefined' or 'null'.
    at getAuthHeader (/Users/awm/dev/dhis2/cli/d2/packages/utils/src/cmds/schema/index.js:70:34)
    at schemasFromUrl (/Users/awm/dev/dhis2/cli/d2/packages/utils/src/cmds/schema/index.js:95:47)
    at getSchemas (/Users/awm/dev/dhis2/cli/d2/packages/utils/src/cmds/schema/diff/index.js:47:25)
    at Object.run [as handler] (/Users/awm/dev/dhis2/cli/d2/packages/utils/src/cmds/schema/diff/index.js:188:31)
    at Object.runCommand (/Users/awm/dev/dhis2/cli/d2/node_modules/yargs/lib/command.js:242:26)
    at Object.parseArgs [as _parseArgs] (/Users/awm/dev/dhis2/cli/d2/node_modules/yargs/yargs.js:1078:30)
    at Object.runCommand (/Users/awm/dev/dhis2/cli/d2/node_modules/yargs/lib/command.js:200:96)
    at Object.parseArgs [as _parseArgs] (/Users/awm/dev/dhis2/cli/d2/node_modules/yargs/yargs.js:1078:30)
    at Object.parse (/Users/awm/dev/dhis2/cli/d2/node_modules/yargs/yargs.js:542:19)
    at module.exports (/Users/awm/dev/dhis2/cli/d2/node_modules/@dhis2/cli-helpers-engine/lib/makeEntryPoint.js:43:11)
```

This comes originally from `diff/index.js:190`.  I think some more testing of un-configured schema diffing might be in order!

Birkbjo: Sorry, I forgot to fallback to prompt if no auth was given. Should be fixed now.

Issue #45, opened 2019-05-13
Single port cluster 


varl:
One question which may be important - do we want to support arbitrary context paths per cluster, or would it be sufficient to support a singular "cluster" gateway which proxies to multiple "versioned and named" clusters behind it.  I.E.

```
d2 cluster config --port 8080
d2 cluster add --version dev --name mydevinstance
d2 cluster add --version 2.32 --name latest
```
(or something better with nicer commands)

Could result in a single gateway pointing at multiple clusters behind-the-scenes, so you could access `localhost:8080/mydevinstance` and `localhost:8080/latest`.  This is a little less flexible but a little more controlled than allowing arbitrary context specification.

_Originally posted by @amcgee in https://github.com/dhis2/cli/pull/39#issuecomment-490834985_

PR #46, opened 2019-05-14
feat: decouple tag, name, and version

varl:
In preparation for #45 this PR experiments with the consequences of decoupling the version, tag, and name of a cluster.

### DIFF
--- a/packages/cluster/src/commands/up.js
+++ b/packages/cluster/src/commands/up.js
@@ -8,7 +8,16 @@ const {
 } = require('../common')
 const { seed: doSeed } = require('../db')
 
-const run = async function({ v, port, seed, seedFile, update, tag, ...argv }) {
+const run = async function({
+    name,
+    port,
+    seed,
+    seedFile,
+    update,
+    tag,
+    version,
+    ...argv
+}) {
     const cacheLocation = await initDockerComposeCache({
         cache: argv.getCache(),
         dockerComposeRepository: argv.cluster.dockerComposeRepository,
@@ -21,22 +30,26 @@ const run = async function({ v, port, seed, seedFile, update, tag, ...argv }) {
     }
 
     if (seed || seedFile) {
-        await doSeed({ cacheLocation, v, path: seedFile, update, ...argv })
+        const resolvedVersion = version ? version : name
+        await doSeed({
+            cacheLocation,
+            resolvedVersion,
+            path: seedFile,
+            update,
+            ...argv,
+        })
     }
 
-    let image = makeDockerImage(v)
-    if (tag) {
-        image = tag
-    }
+    const image = tag ? tag : makeDockerImage(version)
 
-    reporter.info(`Spinning up cluster version ${chalk.cyan(v)}`)
+    reporter.info(`Spinning up cluster ${chalk.cyan(name)}`)
     const res = await tryCatchAsync(
         'exec(docker-compose)',
         exec({
             cmd: 'docker-compose',
             args: [
                 '-p',
-                makeComposeProject(v),
+                makeComposeProject(name),
                 '-f',
                 path.join(cacheLocation, 'docker-compose.yml'),
                 'up',
@@ -44,7 +57,8 @@ const run = async function({ v, port, seed, seedFile, update, tag, ...argv }) {
             ],
             env: {
                 DHIS2_CORE_TAG: image,
-                DHIS2_CORE_VERSION: v,
+                DHIS2_CORE_VERSION: version,
+                DHIS2_CORE_NAME: name,
                 DHIS2_CORE_PORT: port,
             },
             pipe: true,
@@ -57,7 +71,7 @@ const run = async function({ v, port, seed, seedFile, update, tag, ...argv }) {
 }
 
 module.exports = {
-    command: 'up <v>',
+    command: 'up <name>',
     desc: 'Spin up a new cluster',
     aliases: 'u',
     builder: {
@@ -87,6 +101,12 @@ module.exports = {
         tag: {
             alias: 't',
             desc: 'Use the specified tag',
+            type: 'string',
+            default: '',
+        },
+        version: {
+            alias: 'v',
+            desc: 'Set the version',
             type: 'string',
             default: '',
         },
--- a/packages/cluster/src/commands/up.js
+++ b/packages/cluster/src/commands/up.js
@@ -125,6 +125,7 @@ module.exports = {
             default: '',
         },
         root: {
+            alias: 'r',
             desc: 'Serve on the root context path',
             type: 'boolean',
             default: false,
--- a/packages/cluster/src/commands/up.js
+++ b/packages/cluster/src/commands/up.js
@@ -19,7 +19,14 @@ const run = async function({
     ver,
     ...argv
 }) {
-    console.info(argv)
+    const { cluster } = argv
+
+    const resolvedVersion = ver ? ver : name
+    const resolvedTag = tag
+        ? tag
+        : cluster.tag.replace(/{version}/g, resolvedVersion)
+    const resolvedRepo = repo ? repo : cluster.repo
+    const resolvedPort = port ? port : cluster.port
 
     const cacheLocation = await initDockerComposeCache({
         cache: argv.getCache(),
@@ -35,7 +42,7 @@ const run = async function({
     if (seed || seedFile) {
         await doSeed({
             cacheLocation,
-            ver: ver || argv.cluster.ver,
+            dbVersion: resolvedVersion,
             name,
             path: seedFile,
             update,
@@ -58,11 +65,11 @@ const run = async function({
                 '-d',
             ],
             env: {
-                DHIS2_CORE_REPO: repo,
-                DHIS2_CORE_TAG: tag || argv.cluster.tag,
-                DHIS2_CORE_VERSION: ver || argv.cluster.ver,
                 DHIS2_CORE_NAME: name,
-                DHIS2_CORE_PORT: port || argv.cluster.port,
+                DHIS2_CORE_REPO: resolvedRepo,
+                DHIS2_CORE_TAG: resolvedTag,
+                DHIS2_CORE_VERSION: resolvedVersion,
+                DHIS2_CORE_PORT: resolvedPort,
             },
             pipe: true,
         })
--- a/packages/cluster/src/commands/up.js
+++ b/packages/cluster/src/commands/up.js
@@ -19,6 +19,7 @@ const run = async function({
     update,
     image,
     dhis2Version,
+    root,
     ...argv
 }) {
     const { cluster } = argv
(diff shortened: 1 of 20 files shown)


varl:
This decouples the name of the cluster from the version and image it is running.

The `name` is the name of the cluster, and the context path where the application will be accessible.

`repo` and `tag` refers to the Docker repo and tag.

```bash
# run a specific docker repo and tag on the customdev context
d2 cluster up customdev --repo amcgee/dhis2-core --tag dev-latest
# => http://localhost:8080/customdev



# set the DHIS2 version to use to find the right DB version dump
d2 cluster up dev --ver=dev --seed
# => http://localhost:8080/dev



# use defaults from `~/.config/d2` config
```
cat ~/.config/d2/config.js

module.exports = {
    cluster: {
        dockerComposeRepository: 'https://github.com/varl/dhis2-backend/archive/master.tar.gz',
        demoDatabaseURL: 'https://github.com/dhis2/dhis2-demo-db/blob/master/sierra-leone/{version}/dhis2-db-sierra-leone.sql.gz?raw=true',
        tag: 'dev-latest-alpine',
        repo: 'dhis2/core',
        ver: 'dev',
    }
}

d2 cluster up mydev
# => http://localhost:8080/mydev
```

Still a work in progress.

**Comment by varl** [comment]
Depends on https://github.com/varl/dhis2-backend as there are some `docker-compose.yml` changes needed to be able to decouple everything.

**Comment by varl** [comment]
New config file:

```js
module.exports = {
    cluster: {
        dockerComposeRepository: 'https://github.com/varl/dhis2-backend/archive/master.tar.gz',
        demoDatabaseURL: 'https://github.com/dhis2/dhis2-demo-db/blob/master/sierra-leone/{version}/dhis2-db-sierra-leone.sql.gz?raw=true',
        tag: '{version}-latest-alpine',
        repo: 'dhis2/core',
    }
}
```

This works, too:

`d2 cluster up adev --repo amcgee/dhis2-core --tag dev-alpine --seed --ver 2.32`

Spins up amcgee/dhis2-core:dev-alpine with a 2.32 database and deploys it on `/adev`

if `--ver` is not passed, it falls back to try to use the `name` ("adev") as the version. This works if it's `dev`, `2.32`, or something logical. But if a completely custom is used the `--ver` is recommended.

If the `--tag` arg is passed, it is used, or it reads the default from the config file and replaces the `{version}` marker with the resolved version from above.

So `d2 cluster up 2.32 --seed` and `d2 cluster up dev` will do the right thing.

varl:
One thing to look out for is that the default config for the entire cli lives in cli-helpers-engine.

varl:
`--version` is shadowed by the global switch for `d2`, thinking of renaming `--ver` to `--dhis2-version` since that is what we are looking for here.

varl:
This works now. :)

varl:
@amcgee This does not aim to add the API proposed in #45, nor add the single-port gateway, but do the groundwork to decouple some assumptions about Docker repo/tag, DHIS2 versions, context, etc.

Requires this line to be changed: https://github.com/amcgee/dhis2-backend/blob/master/docker-compose.yml#L4 to:

```
    image: "${DHIS2_CORE_REPO}:${DHIS2_CORE_TAG}"
```

And either a `~/.config/d2/config.js` file needs to exist with the new options, or all args needs to be passed in until the defaults are updated here: https://github.com/dhis2/cli-helpers-engine/blob/master/lib/configDefaults.js#L38-L43 with:

```
        tag: '{version}-latest-alpine',
        repo: 'dhis2/core',
```

Would be nice to be able to define the default configuration for the `cluster` command in the `cluster` command module, but haven't looked into that yet.

Mohammer5:
Doesn't really matter, but you could change this:

```js
    const resolvedVersion = ver ? ver : name
    const resolvedTag = tag
        ? tag
        : cluster.tag.replace(/{version}/g, resolvedVersion)
    const resolvedRepo = repo ? repo : cluster.repo
    const resolvedPort = port ? port : cluster.port
```

into this

```js
    const resolvedVersion = ver || name
    const resolvedTag = tag || cluster.tag.replace(/{version}/g, resolvedVersion)
    const resolvedRepo = repo || cluster.repo
    const resolvedPort = port || cluster.port
```

Reads a bit nicer, but I don't care, so totally up to you

amcgee:
@varl very nice!

I agree it would be awesome to support per-module config defaults.  I'll look into adding that to the engine.  In the meantime maybe it's best to just define the defaults in-line here, following on what @Mohammer5 suggested - `const resolvedRepo = repo || cluster.repo || defaultRepo` would work for that and not require the defaults to be in the `engine` itself.  I think that's a nice first step towards supporting per-cluster defaults.

amcgee:
Is there a reason to configure `repo` and `tag` separately?  Since we support version-substitution we could just pass an `image` (or some other name) parameter like `dhis2/core:{version}-alpine` and not bother with the extra parameters?  Just a thought that might simplify this a little.

varl:
Thanks for the input, cleaned it up a bit. 👍

There's more to do but with the `defaults` strategy we could remove the defaults from the `cli-helpers-engine` here: https://github.com/dhis2/cli-helpers-engine/blob/master/lib/configDefaults.js#L35-L44

So the responsibility of the `cli-helpers-engine` would be to load the user's configuration if available, and the default configuration will be handled by the respective command module, demonstrated in `defaults.js`.

Those defaults can be reused for the default values for e.g. port, so there is a single place to change them if need be.

varl:
One heads up here; this does not support hosting the context from `/` at the moment, it will always serve DHIS2 from the `name` (which is always the name of the cluster). Personally I think this makes sense.

I can add a `--context-path` option to the command though where one could override and split the cluster name from the context path. It would introduce another env variable which needs to be used in the docker-compose repo.

varl:
8c4b70a adds a `--root` option to allow for a root context deploy, e.g. http://localhost:8080.

`d2 cluster up dev --root`

varl:
@amcgee I'm pretty much content with this now, ready for review.

varl:
This line: https://github.com/dhis2/cli/pull/46/files#diff-daa1df464973ea1a08c2c122f6a38b65R3

Depends on the outcome of what the image pushed to hub.docker.com/r/dhis2/core is, which will be decided by: https://github.com/dhis2/dhis2-core/pull/3326/files#diff-48b252da11ed2edeae1188be3f75660eR32

varl:
This brings up the question if we should merge `amcgee/dhis2-backend` to `dhis2/docker-compose` or something like that, then I can submit a PR for my changes from `varl/dhis2-backend`.

That is if this PR is the way to move forward for now.

packages/cluster/src/commands/up.js
```diff
@@ -5,13 +5,40 @@ const {
...
+
+    const resolvedVersion = dhis2Version || name
+    const resolvedImage = substituteVersion(
+        image || cluster.image || defaults.image,
+        resolvedVersion
+    )
+
+    const resolvedPort = port || cluster.port || defaults.port
+    const resolvedRoot = root || cluster.root || defaults.root
+    const resolvedContextPath = resolvedRoot ? '' : `/${name}`
```
amcgee: I like the simplicity of this but I worry about it.  It's a breaking change to default to a non-root context.  Should we default to the root context and have a `--no-root` flag (or something better) to switch to deploying on the probably less-common non-root context?

packages/cluster/src/defaults.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+    dhis2Version: '',
+    image: 'dhis2/core:{version}-latest-alpine',
+    port: 8080,
+    root: false,
```
amcgee: If this were `true` it wouldn't be a breaking change

packages/cluster/src/defaults.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+    dhis2Version: '',
+    image: 'dhis2/core:{version}-latest-alpine',
```
amcgee: Actually this will work (for `dev` and `2.32`) currently since they're currently set up to build automatically on docker hub.  Released versions (`2.32.0-alpine`) won't work though

amcgee: If this were `dhis2/core:{version}-alpine` it would work as `d2 cluster up 2.32-latest` and `d2 cluster up 2.32.0`, however it's a change from the previous behavior which would support `d2 cluster up 2.32` and `d2 cluster up 2.32.0`.  I think that's an OK change, but will require a major version bump I think

amcgee: Alternatively we could transform `2.32` into `2.32-latest` (also for dev and other branches) in the `up` code.

Also worth noting - `2.32-latest` won't find a db to seed, currently, but existing behavior already fails for `2.32.0`

varl: > Actually this will work (for `dev` and `2.32`) currently since they're currently set up to build automatically on docker hub. Released versions (`2.32.0-alpine`) won't work though

Yeah, in that case you would have to pass in a full image reference: `--image dhis2/core:2.32.0-alpine`.

I want to let https://github.com/dhis2/notes/issues/48 get settled, then we can figure out the patterns to support.

packages/cluster/src/defaults.js
varl: That's true. :thinking:

packages/cluster/src/commands/up.js
varl: Ok, I yield. The implementation docs uses the ROOT as well so let's go with that: https://docs.dhis2.org/2.32/en/implementer/html/install_server_setup.html#install_tomcat_dhis2_installation

varl: Rename it to `--custom-context`?

packages/cluster/src/defaults.js
varl: Reverted to use amcgee/dhis2-core for now. Will tackle the official images in another PR.

varl:
This is now compatible with `master` of d2-cluster.

amcgee:
Found an edge case we might want to deal with or call a "known issue":

```
> d2 cluster up testing --dhis2-version 2.32 --seed
> d2 cluster down testing
> d2 cluster up testing
```

The last `up` doesn't have the context to know that it should pass a 2.32 image name, so it will break because `amcgee/dhis2-core:testing-alpine` doesn't exist.  By contrast, when the name is used as the version this works.

To solve this the right way we'll need to keep state about all named containers and their parameters.  We probably want to do that anyway, but I'm OK with calling that a future enhancement.

varl:
> To solve this the right way we'll need to keep state about all named containers and their parameters. We probably want to do that anyway, but I'm OK with calling that a future enhancement.

Ah, shoot. I forgot about this. I ran into the same problem, put it on the back burner and subsequently forgot about it.

I was thinking about wanting to be able to e.g. configure my clusters in the config file:

```
{
  cluster: {
    instances: {
      '{name}': {
        customContext: true,
        dhis2Version: '2.32',
      }
    }
  }
}
```

And that it should interact with the cache for un-configured values, which were passed at the time of the first `up` command.

varl:
> if we're OK having the above known issue I'm happy with this change!

Yeah, I think so.

PR #49, opened 2019-05-14
fix: update @dhis2/cli style 3.2.1

varl:


amcgee:
Should we bump `cli-helpers-engine` in `cli-style` before upgrading here?

varl:
@amcgee It's at latest already.

https://github.com/dhis2/cli-helpers-engine/blob/master/package.json#L3
https://www.npmjs.com/package/@dhis2/cli-helpers-engine

amcgee:
Ah you're right, there's still a `1.0.1` in here for some reason but probably OK

PR #53, opened 2019-05-23
Decouple configs from cluster

varl:
Creates a new Cache for `d2-cluster` based on the global cache.

Would be nice to pair with https://github.com/dhis2/cli-helpers-engine/pull/15/files to remove the layered cache directory.

### DIFF
--- a/packages/cluster/src/commands/down.js
+++ b/packages/cluster/src/commands/down.js
@@ -4,21 +4,19 @@ const { exec, reporter } = require('@dhis2/cli-helpers-engine')
 const {
     initDockerComposeCache,
     resolveConfiguration,
+    makeEnvironment,
+    cleanCache,
 } = require('../common')
 
-const defaults = require('../defaults')
-
-const run = async function({ name, clean, getCache, cluster, ...argv }) {
-    const {
-        dockerComposeRepository,
-        dockerComposeDirectory,
-    } = resolveConfiguration(argv, {}, cluster)
+const run = async function(argv) {
+    const { name, clean, getCache } = argv
+    const cfg = await resolveConfiguration(argv)
 
     const cacheLocation = await initDockerComposeCache({
-        name,
+        composeProjectName: name,
         cache: getCache(),
-        dockerComposeRepository,
-        dockerComposeDirectory,
+        dockerComposeRepository: cfg.dockerComposeRepository,
+        dockerComposeDirectory: cfg.dockerComposeDirectory,
         force: false,
     })
 
@@ -27,7 +25,7 @@ const run = async function({ name, clean, getCache, cluster, ...argv }) {
         process.exit(1)
     }
 
-    console.log(`Winding down cluster ${chalk.cyan(name)}`)
+    reporter.info(`Winding down cluster ${chalk.cyan(name)}`)
     try {
         await exec({
             cmd: 'docker-compose',
@@ -38,15 +36,11 @@ const run = async function({ name, clean, getCache, cluster, ...argv }) {
                 path.join(cacheLocation, 'docker-compose.yml'),
                 'down',
             ].concat(clean ? ['--volumes'] : []),
-            env: {
-                DHIS2_CORE_IMAGE: defaults.image,
-                DHIS2_CORE_NAME: name,
-                DHIS2_CORE_PORT: defaults.port,
-            },
+            env: makeEnvironment(cfg),
         })
 
         if (clean) {
-            await cleanCache(getCache(), name)
+            cleanCache(argv)
         }
     } catch (e) {
         reporter.error('Failed to execute docker-compose', e)
--- a/packages/cluster/src/commands/down.js
+++ b/packages/cluster/src/commands/down.js
@@ -4,6 +4,7 @@ const { exec, reporter } = require('@dhis2/cli-helpers-engine')
 const {
     initDockerComposeCache,
     resolveConfiguration,
+    makeComposeProject,
     makeEnvironment,
     cleanCache,
 } = require('../common')
@@ -31,7 +32,7 @@ const run = async function(argv) {
             cmd: 'docker-compose',
             args: [
                 '-p',
-                name,
+                makeComposeProject(name),
                 '-f',
                 path.join(cacheLocation, 'docker-compose.yml'),
                 'down',
--- a/packages/cluster/src/commands/down.js
+++ b/packages/cluster/src/commands/down.js
@@ -3,7 +3,6 @@ const path = require('path')
 const { exec, reporter } = require('@dhis2/cli-helpers-engine')
 const {
     initDockerComposeCache,
-    makeComposeProject,
     resolveConfiguration,
 } = require('../common')
 
@@ -15,9 +14,8 @@ const run = async function({ name, clean, getCache, cluster, ...argv }) {
         dockerComposeDirectory,
     } = resolveConfiguration(argv, {}, cluster)
 
-    const composeProjectName = makeComposeProject(name)
     const cacheLocation = await initDockerComposeCache({
-        composeProjectName,
+        name,
         cache: getCache(),
         dockerComposeRepository,
         dockerComposeDirectory,
@@ -35,7 +33,7 @@ const run = async function({ name, clean, getCache, cluster, ...argv }) {
             cmd: 'docker-compose',
             args: [
                 '-p',
-                composeProjectName,
+                name,
                 '-f',
                 path.join(cacheLocation, 'docker-compose.yml'),
                 'down',
@@ -46,6 +44,10 @@ const run = async function({ name, clean, getCache, cluster, ...argv }) {
                 DHIS2_CORE_PORT: defaults.port,
             },
         })
+
+        if (clean) {
+            await cleanCache(getCache(), name)
+        }
     } catch (e) {
         reporter.error('Failed to execute docker-compose', e)
         process.exit(1)
--- a/packages/cluster/src/commands/down.js
+++ b/packages/cluster/src/commands/down.js
@@ -14,7 +14,10 @@ const run = async function({ name, clean, getCache, cluster, ...argv }) {
         dockerComposeRepository,
         dockerComposeDirectory,
     } = resolveConfiguration(argv, {}, cluster)
+
+    const composeProjectName = makeComposeProject(name)
     const cacheLocation = await initDockerComposeCache({
+        composeProjectName,
         cache: getCache(),
         dockerComposeRepository,
         dockerComposeDirectory,
@@ -32,7 +35,7 @@ const run = async function({ name, clean, getCache, cluster, ...argv }) {
             cmd: 'docker-compose',
             args: [
                 '-p',
-                makeComposeProject(name),
+                composeProjectName,
                 '-f',
                 path.join(cacheLocation, 'docker-compose.yml'),
                 'down',
--- a/packages/cluster/src/commands/down.js
+++ b/packages/cluster/src/commands/down.js
@@ -40,7 +40,10 @@ const run = async function(argv) {
         })
 
         if (clean) {
-            cleanCache(argv)
+            cleanCache({
+                name,
+                cache: getCache(),
+            })
         }
     } catch (e) {
         reporter.error('Failed to execute docker-compose', e)
--- a/packages/cluster/src/commands/logs.js
+++ b/packages/cluster/src/commands/logs.js
@@ -3,6 +3,7 @@ const path = require('path')
 const { reporter, exec, tryCatchAsync } = require('@dhis2/cli-helpers-engine')
 const {
     initDockerComposeCache,
+    makeComposeProject,
     makeEnvironment,
     resolveConfiguration,
 } = require('../common')
@@ -36,7 +37,7 @@ const run = async function(argv) {
             cmd: 'docker-compose',
             args: [
                 '-p',
-                name,
+                makeComposeProject(name),
                 '-f',
                 path.join(cacheLocation, 'docker-compose.yml'),
                 'logs',
--- a/packages/cluster/src/commands/logs.js
+++ b/packages/cluster/src/commands/logs.js
@@ -1,11 +1,7 @@
 const chalk = require('chalk')
 const path = require('path')
 const { reporter, exec, tryCatchAsync } = require('@dhis2/cli-helpers-engine')
-const {
-    initDockerComposeCache,
-    makeComposeProject,
-    resolveConfiguration,
-} = require('../common')
+const { initDockerComposeCache, resolveConfiguration } = require('../common')
 const defaults = require('../defaults')
 
 const run = async function({ service, name, cluster, ...argv }) {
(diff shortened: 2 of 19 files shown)


packages/cluster/src/common.js
```diff
@@ -1,23 +1,32 @@
...
 
 module.exports.initDockerComposeCache = async ({
+    composeProjectName,
     cache,
     dockerComposeRepository,
     force,
 }) => {
-    const exists = await cache.exists(dockerComposeCacheName)
+    // make new cache dir per cluster in the root cache
+    const clusterCache = new Cache({
```
amcgee: It should be possible to cache and lookup files with deep paths using the main cache, actually, so I don't think this multi-layered cache is necessary?  You should be able to just do `cache.exists(path.join(composeProjectName, dockerComposeCacheName))`

packages/cluster/src/common.js
```diff
@@ -32,8 +41,29 @@ module.exports.initDockerComposeCache = async ({
     return null
 }
 
-module.exports.makeComposeProject = version => `d2-cluster-${version}`
-module.exports.makeDockerImage = version => `${version}-alpine`
+module.exports.writeCache = (cache, loc) => {
+    try {
+        const target = path.dirname(loc)
+        fs.writeFileSync(
+            path.join(target, 'clusterCache.json'),
```
amcgee: Probably good to factor the filename out as a const

packages/cluster/src/common.js
varl: I'll give it a whack.

packages/cluster/src/common.js
varl: Will do. ⚡️

varl: b727e65

packages/cluster/src/common.js
varl: The reason I'm using the multi-layer cache is actually to get it to create the sub-directory for me. Just `cache.exists()` just checks for it, it doesn't guarantee that it is there.

varl: What we have now:
```
ls -1 ~/.cache/d2/cache/
cluster-db-dev.sql.gz
d2-cluster-docker-compose-v2
```

I was going for something like:

- ~/.cache/d2/cache/{cluster name}/docker-compose/docker-compose.yml`
- ~/.cache/d2/cache/{cluster name}/clusterConfig.json`

varl:
Rebased on `master`, variation on the multi-layered `Cache`.

Produces this structure:

```
tree ~/.cache/d2/ -L 4
/home/varl/.cache/d2/
└── cache
    ├── cluster
    │   ├── d2-cluster-dev
    │   │   └── docker-compose
    │   └── d2-cluster-master
    │       └── docker-compose
    ├── cluster-db-dev.sql.gz
    └── cluster-db-master.sql.gz
```

Depends on https://github.com/dhis2/cli-helpers-engine/pull/16

varl:
```
varl@veronika:~/dev/dhis2/libs/cli$ ./packages/main/bin/d2 debug cache list
┌──────────────────────────┬──────────┬─────────────────────┐
│ Name                     │ Size     │ Modified            │
├──────────────────────────┼──────────┼─────────────────────┤
│ cluster                  │ 512      │ 2019-05-27 17:50:36 │
├──────────────────────────┼──────────┼─────────────────────┤
│ cluster-db-dev.sql.gz    │ 98904634 │ 2019-05-27 17:50:42 │
├──────────────────────────┼──────────┼─────────────────────┤
│ cluster-db-master.sql.gz │ 165532   │ 2019-05-27 17:50:20 │
└──────────────────────────┴──────────┴─────────────────────┘
varl@veronika:~/dev/dhis2/libs/cli$ ./packages/main/bin/d2 debug cache list cluster
┌───────────────────┬──────┬─────────────────────┐
│ Name              │ Size │ Modified            │
├───────────────────┼──────┼─────────────────────┤
│ d2-cluster-dev    │ 512  │ 2019-05-27 17:50:37 │
├───────────────────┼──────┼─────────────────────┤
│ d2-cluster-master │ 512  │ 2019-05-27 17:44:02 │
└───────────────────┴──────┴─────────────────────┘
varl@veronika:~/dev/dhis2/libs/cli$ ./packages/main/bin/d2 debug cache list cluster d2-cluster-dev
┌───────────────────┬──────┬─────────────────────┐
│ Name              │ Size │ Modified            │
├───────────────────┼──────┼─────────────────────┤
│ d2-cluster-dev    │ 512  │ 2019-05-27 17:50:37 │
├───────────────────┼──────┼─────────────────────┤
│ d2-cluster-master │ 512  │ 2019-05-27 17:44:02 │
└───────────────────┴──────┴─────────────────────┘
```

I think I'll rename the `d2-cluster-dev` to just `dev` since it's already subdired.

packages/cluster/src/commands/down.js
```diff
@@ -22,23 +25,23 @@ const run = async function({ name, clean, getCache, cluster, ...argv }) {
...
-            env: {
-                DHIS2_CORE_IMAGE: defaults.image,
-                DHIS2_CORE_NAME: name,
-                DHIS2_CORE_PORT: defaults.port,
-            },
+            env: makeEnvironment(cfg),
         })
+
+        if (clean) {
+            cleanCache(argv)
```
varl: I think it makes sense to purge the cache for the cluster when we run `--clean`.

packages/cluster/src/common.js
```diff
@@ -1,41 +1,48 @@
-const { reporter } = require('@dhis2/cli-helpers-engine')
+const path = require('path')
+const fs = require('fs')
+
+const { Cache, reporter } = require('@dhis2/cli-helpers-engine')
 
 const defaults = require('./defaults')
 
-const dockerComposeCacheName = 'd2-cluster-docker-compose-v2'
+const clusterDir = 'cluster'
```
varl: This could be inferred from `argv` since we have the command name. It does make it less obvious so went with readability++ for now. If we pull these out to "common for all of d2" it would make more sense to infer it.

packages/cluster/src/common.js
```diff
@@ -69,8 +76,13 @@ function replacer(string, token, value) {
     return string.replace(regexp, value)
 }
 
-function resolveConfiguration(argv = {}, cache = {}, config = {}) {
-    const resolved = Object.assign({}, defaults, config, cache, argv)
+async function resolveConfiguration(argv = {}) {
+    const file = path.join(clusterDir, argv.name, cacheFile)
+    const cache = JSON.parse(await argv.getCache().read(file))
```
varl: `read` returns `null` if the file doesn't exist/cannot be read which works with `JSON.parse`. Could make this more verbose and let the `Cache.read` function throw an error instead.

packages/cluster/src/common.js
```diff
@@ -88,9 +100,28 @@ function resolveConfiguration(argv = {}, cache = {}, config = {}) {
 
     reporter.debug('Resolved configuration\n', resolved)
 
+    await argv.getCache().write(
```
varl: Same deal here, the `Cache.write` just logs an error and continues if the cache write fails. I don't think that it's such a big deal if the cache fails to be written, and I want the behaviours of the `read`/`write` pair to match, so if one throws, the other should throw. Since the implementation of `read` returns `null` if it fails, I let this one log & continue on a cache write fail.

varl:
Updated https://github.com/dhis2/cli-helpers-engine/pull/16 with a `read`/`write` implementation for the cache, and could simplify this PR as a result.

packages/cluster/src/commands/down.js
varl: Could remove this and users could still do it manually with `d2 debug cache purge cluster/{name}` though. Not sure if I expect the configuration to be saved when I run `d2 cluster down {name} --clean` or not..

varl:
@amcgee I think this is ready for review and testing without using `link`s.

packages/cluster/src/commands/down.js
varl: Since we support cached and persistent configurations, I think it makes sense to clear the cache when doing `--clean`.

amcgee:
@varl looks like there's been a bunch of activity since you said this was ready for review, are you still iterating on it?

varl:
@amcgee No, I just rebased on master to bring it up to date. Looks like there's another conflict so I'll do it once more. :innocent:

varl:
Rebased on `master` so incorporates #61.

varl:
@amcgee This is ready for review.

packages/cluster/src/commands/up.js
```diff
@@ -48,7 +49,7 @@ const run = async function(argv) {
             cmd: 'docker-compose',
             args: [
                 '-p',
-                makeComposeProject(name),
+                name,
```
amcgee: I think it's important to keep the docker-compose project namespaced (i.e. prefixed with `d2-cluster` as it was previously), otherwise `dev` is very generic and could well mess up the `docker ps` call below (if there's any other docker container on the maching with `dev` in the name.

packages/cluster/src/common.js
```diff
@@ -80,8 +85,23 @@ function replacer(string, token, value) {
     return string.replace(regexp, value)
 }
 
-function resolveConfiguration(argv = {}, cache = {}, config = {}) {
-    const resolved = Object.assign({}, defaults, config, cache, argv)
+async function resolveConfiguration(argv = {}) {
+    const file = path.join(clusterDir, argv.name, cacheFile)
+    const currentCache = JSON.parse(await argv.getCache().read(file))
```
amcgee: This should be try-guarded or at least have a `fs.exists` check

packages/cluster/src/common.js
```diff
@@ -99,9 +119,28 @@ function resolveConfiguration(argv = {}, cache = {}, config = {}) {
...
+            null,
+            4
+        )
+    )
+
     return resolved
 }
 
+module.exports.cleanCache = async argv =>
+    await argv.getCache().purge(path.join(clusterDir, argv.name))
```
amcgee: Why doesn't this use the `({ name })` object parameter deconstruction form?  I find `argv` arguments opaque unless there are potentially unknown or dynamic properties.

packages/cluster/src/commands/status.js
```diff
@@ -1,11 +1,10 @@
 const { exec, reporter } = require('@dhis2/cli-helpers-engine')
-const { makeComposeProject } = require('../common')
 
 const run = async function({ name, ...argv }) {
     try {
         await exec({
             cmd: 'docker',
-            args: ['ps', '--filter', `name=${makeComposeProject(name)}`],
+            args: ['ps', '--filter', `name=${name}`],
```
amcgee: This will return extra, unrelated information if we don't prefix the compose project name.

packages/cluster/src/common.js
varl: 4f7d7d2

packages/cluster/src/common.js
varl: I thought about it, but I thought the other way around, that it makes it less clear what is being used from where.

Changed in 4f7d7d2

packages/cluster/src/commands/up.js
varl: Ok, will change so the docker-compose parts are prefixed.

varl: 88e7702

packages/cluster/src/commands/status.js
varl: Done.

varl:
@amcgee changed the code with regards to your comments.

packages/cluster/src/commands/seed.js
```diff
@@ -2,20 +2,16 @@ const { reporter } = require('@dhis2/cli-helpers-engine')
 const { initDockerComposeCache, resolveConfiguration } = require('../common')
 const { seed } = require('../db')
 
-const defaults = require('../defaults')
+const run = async function(argv) {
+    const { name, getCache } = argv
```
amcgee: Should this (`name`, particularly) come from `cfg` not directly from `argv`?

packages/cluster/src/db.js
```diff
@@ -1,4 +1,4 @@
-const { makeComposeProject, substituteVersion } = require('./common')
+const { substituteVersion, makeComposeProject } = require('./common')
```
amcgee: No-op change, might be worth reverting

packages/cluster/src/commands/logs.js
```diff
@@ -4,30 +4,33 @@ const { reporter, exec, tryCatchAsync } = require('@dhis2/cli-helpers-engine')
...
-    const {
-        dockerComposeRepository,
-        dockerComposeDirectory,
-    } = resolveConfiguration(argv, {}, cluster)
+const run = async function(argv) {
+    const { service, name } = argv
+    const cfg = await resolveConfiguration(argv)
+
     const cacheLocation = await initDockerComposeCache({
+        composeProjectName: name,
```
amcgee: Should this also be `makeComposeProject(name)` or the parameter just called `name`?  Otherwise maybe `initDockerComposeCache::composeProjectName` is a bit misleading...

varl: Yeah, I've considered that too.

The main reason I want `name` here is to make the cache a lot easier to browse: `d2 debug cache list clusters/dev/docker-compose`

If we use `makeComposeProject`: `d2 debug cache list clusters/d2-cluster-dev/docker-compose`

Since all the compose projects are already namespaced to `clusters/{name}` I don't think it makes sense to prefix it again with `d2-cluster-{name}`.

It's still the cache directory for the compose project, so the function name names sense (`initDockerComposeCache`), it's just the argument that is slightly misleading. I can rename it to `composeCacheName` if that makes it clearer.

packages/cluster/src/db.js
varl: eaccd9c

packages/cluster/src/commands/seed.js
varl: Hm, `cfg` doesn't actually have the `name` since it is always passed in as an argument, that's why I always grab it from `argv`.

packages/cluster/src/commands/logs.js
varl: @amcgee Alternatively `initDockerComposeCache::cacheDirName`?

packages/cluster/src/commands/seed.js
amcgee: I think `cfg` does have it, since internally it does an `Object.assign(..., argv)` but I might be wrong.  This is fine as-is!

packages/cluster/src/commands/logs.js
amcgee: We can rename this later but I think `name` might even be fine

packages/cluster/src/common.js
```diff
@@ -80,8 +85,29 @@ function replacer(string, token, value) {
...
-function resolveConfiguration(argv = {}, cache = {}, config = {}) {
-    const resolved = Object.assign({}, defaults, config, cache, argv)
+async function resolveConfiguration(argv = {}) {
+    const file = path.join(clusterDir, argv.name, cacheFile)
+
+    let currentCache
+    try {
+        currentCache = JSON.parse(await argv.getCache().read(file))
+    } catch (e) {
+        log.debug('JSON parse of cache file failed', e)
```
amcgee: `log` is undefined here!

amcgee: I noticed this testing the "upgrade" path where an previously-existing cluster is managed (i.e.`d2 cluster down dev --clean`) with the latest cli

amcgee: (went ahead and fixed this myself)

varl:
No worries, the PR is not that big but the implications of it are huge so good to be thorough!

packages/cluster/src/common.js
varl: Thanks!

PR #54, opened 2019-05-24
feat: support official docker images

varl:
Started with the `makeDockerImage` resolutions for supporting channels and versions from https://hub.docker.com/u/dhis2.

- Adds support for official dhis2 docker images in different channels (`core`, `core-dev`, `core-canary`)
- Adds `--db-version <string>`
- Adds `--variant <array>` 
- Adds `--channel <string>`
- Adds tests for configuration merging and generating docker image names

#### Combos

```bash
# 2.32 works for both dhis2 and db version
d2 cluster up 2.32 \
  --channel canary

# custom images require explicit context
d2 cluster up custom \
  --image amcgee/dhis2-core:dev-alpine \
  --db-version 2.32 \ 
  --seed

# sometimes the dhis2 and db versions do not match
d2 cluster up stable \
  --dhis2-version 2.31.4 \
  --db-version 2.30 \
  --seed

# dhis2 version can be used for db version if it matches
d2 cluster up 232-dev \
  --channel dev \
  --dhis2-version 2.32 \
  --seed
```

#### Save per cluster configuration

To allow for subsequent startups without all the switches we need a mechanism (e.g. https://github.com/dhis2/cli/pull/53) to save the configuration first passed to the `up` command.

So first time the cluster is created it stores the config in e.g. `~/.cache/d2/d2-cluster/dev/configCache.json`

This enables:

```bash
d2 cluster up dev \
  --custom-context \
  --channel dev \
  --dhis2-version master \
  --db-version dev \
  --seed

d2 cluster down dev

d2 cluster up dev
# will use the `configCache`
```

Mentioning it here to note that implementing that is out of scope for this PR, but there are plans to do it.

### DIFF
--- a/packages/cluster/src/commands/up.js
+++ b/packages/cluster/src/commands/up.js
@@ -4,45 +4,29 @@ const { reporter, exec, tryCatchAsync } = require('@dhis2/cli-helpers-engine')
 const {
     initDockerComposeCache,
     makeComposeProject,
-    makeDockerImage,
+    makeEnvironment,
 } = require('../common')
 
 const defaults = require('../defaults')
 const { seed: doSeed } = require('../db')
 
-const run = async function({
-    name,
-    port,
-    seed,
-    seedFile,
-    update,
-    image,
-    dhis2Version,
-    customContext,
-    variant,
-    channel,
-    ...argv
-}) {
-    const { cluster } = argv
+const run = async function(argv) {
+    const {
+        cluster,
+        name,
+        port,
+        seed,
+        seedFile,
+        update,
+        image,
+        dhis2Version,
+        dbVersion,
+        customContext,
+        variant,
+        channel,
+    } = argv
 
-    const resolvedVersion = dhis2Version || cluster.dhis2Version || name
-    const resolvedChannel = channel || cluster.channel || defaults.channel
-    const resolvedImage = image || cluster.image || defaults.image
-    const resolvedVariant = variant || cluster.variant
-
-    const resolvedDockerImage = makeDockerImage(
-        resolvedImage,
-        {
-            channel: resolvedChannel,
-            version: resolvedVersion,
-        },
-        resolvedVariant
-    )
-
-    const resolvedPort = port || cluster.port || defaults.port
-    const resolvedContext =
-        customContext || cluster.customContext || defaults.customContext
-    const resolvedContextPath = resolvedContext ? `/${name}` : ''
+    const runtime = makeEnvironment(argv, {}, cluster)
 
     const cacheLocation = await initDockerComposeCache({
         cache: argv.getCache(),
@@ -60,7 +44,7 @@ const run = async function({
     if (seed || seedFile) {
         await doSeed({
             cacheLocation,
-            dbVersion: resolvedVersion,
+            dbVersion: resolvedDatabaseVersion,
             name,
             path: seedFile,
             update,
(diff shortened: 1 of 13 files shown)


varl:
Adding tests for runtime environment generation.

varl:
There are so many different configuration options that tests are the only solution. Also, yay, tests.

packages/cluster/src/commands/up.js
```diff
@@ -98,13 +85,11 @@ module.exports = {
             alias: 'p',
             desc: 'Specify the port on which to expose the DHIS2 instance',
             type: 'integer',
-            default: defaults.port,
```
varl: If something is specified as "default" here, then it will be set in `argv`, making it hard to override when trying using `argv || cache || config || defaults` as the `argv` will always get set.

packages/cluster/src/commands/up.js
```diff
@@ -4,36 +4,29 @@ const { reporter, exec, tryCatchAsync } = require('@dhis2/cli-helpers-engine')
...
-    const resolvedImage = substituteVersion(
-        image || cluster.image || defaults.image,
-        resolvedVersion
-    )
-
-    const resolvedPort = port || cluster.port || defaults.port
-    const resolvedContext =
-        customContext || cluster.customContext || defaults.customContext
-    const resolvedContextPath = resolvedContext ? `/${name}` : ''
+    const runtime = makeEnvironment(argv, {}, cluster)
```
varl: The second empty object is going to be replaced with the per-cluster cached configuration. The `cluster` object is the default configuration from `~/.config/d2/config.js`.

packages/cluster/tests/substitute-string-keys.js
```diff
@@ -0,0 +1,179 @@
...
+
+    const expected = 'dhis2/core:master-tomcat9-debian-slim'
+
+    t.equal(
+        makeDockerImage(
+            template,
+            {
+                version: 'master',
+            },
+            ['tomcat9', 'debian-slim']
```
amcgee: Question: how does an array like this get passed to the CLI?  Since order is important, how do we enforce that, i.e. linux flavor doesn't get appended before tomcat variant?

packages/cluster/src/commands/up.js
varl: Good idea, I'll do that.

packages/cluster/tests/substitute-string-keys.js
varl: > Question: how does an array like this get passed to the CLI?

There are three options I can see a user trying, two of which works:

1. `--variant tomcat alpine`
2.`--variant tomcat-alpine` 
3. `--variant tomcat, alpine`

**1** and **2** will result in a correct variant being appended to the image `-tomcat-alpine`. The comma separated option (**3**) does not work.

More info about the `.array()` option: https://github.com/yargs/yargs/blob/master/docs/api.md#arraykey

> Since order is important, how do we enforce that, i.e. linux flavor doesn't get appended before tomcat variant?

Then it is better to force the user to option **2** and take the variant as a single slugged string.

varl: Change `variant` to string in aef2f98

packages/cluster/src/commands/up.js
varl: 6979364

varl:
@amcgee did three changes:

1. split the `makeEnvironment` in twain; `resolveConfiguration` and `makeEnvironment`.
2. added default where important (`channel` and `port`) to the description.
3. changed the `--variant` switch to accepting a string, so users would have to say `--variant tomcat-alpine`, this works around the ordering problem with an array switch.

packages/cluster/src/commands/up.js
```diff
@@ -115,24 +88,32 @@ module.exports = {
...
             desc: 'Set the DHIS2 version',
             type: 'string',
-            default: defaults.dhis2Version,
+        },
+        dbVersion: {
+            desc: 'Set the database version',
+            type: 'string',
+        },
+        channel: {
+            desc: 'Set the release channel to use (default: stable)',
```
amcgee: Should this use the same `defaults` logic as the other properties?  So basically that would mean removing the `if !substitutes.channel` check from `makeDockerImage` and adding `channel: 'stable'` to defaults ...?

varl: Sure, I can change the `defaults.channel` to `stable` and use the same logic. 👍 

I cannot remove the `!substitutes.channel` check though since the `makeDockerImage` function doesn't do the configuration resolution itself, and if the function is used with a substitution object which doesn't pass in the `channel`, it won't get replaced with an empty string. Would have been a bit simpler if the stable repo was called `core-stable` and not just `core`, but I'd rather have a little bit of extra complexity here and a nicer repo name than the other way around.

varl: 0929a64

PR #95, opened 2019-07-02
feat: add cluster db commands, update images on up, add compose cmd

amcgee:
This feature PR adds a few handy commands to the `cluster` namespace:

1) It adds `d2 cluster db backup <name> <path>` and `d2 cluster db restore <name> [path]` for easy database lifecycle management outside the cluster volumes.
2) It adds the `d2 cluster compose <name> -- [args...]` command which passes through to the underlying `docker-compose` and allows the user to perform arbitrary `docker-compose` commands without needing to specify the compose project name or the location of the docker-compose.yml file (which is in the d2 cache).  An example usage is `d2 cluster compose dev -- exec db 'sh'` which will execute an interactive shell terminal in the `db` container.
3) It enhances the `d2 cluster up` command by automatically pulling the latest docker images when `--update` is specified

This also deprecates the `d2 cluster seed` command (it's still there to prevent a breaking change, but hidden from the help output) in favor of `d2 cluster db restore`

### DIFF
--- a/packages/cluster/src/commands/compose.js
+++ b/packages/cluster/src/commands/compose.js
@@ -1,0 +1,51 @@
+const path = require('path')
+const { reporter, exec, tryCatchAsync } = require('@dhis2/cli-helpers-engine')
+const {
+    initDockerComposeCache,
+    makeComposeProject,
+    makeEnvironment,
+    resolveConfiguration,
+} = require('../common')
+
+const run = async function(argv) {
+    const { name, service, args } = argv
+    const cfg = await resolveConfiguration(argv)
+
+    const cacheLocation = await initDockerComposeCache({
+        composeProjectName: name,
+        cache: argv.getCache(),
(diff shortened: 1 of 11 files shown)


amcgee:
One question here - does `d2 cluster compose` make sense to pass-through commands to the underlying `docker-compose`?  It might be a bit obtuse, something like `exec` or `run` is a bit better, but is confusing because there's also a concept of `run` or `exec` of a custom command or container within the cluster network (`d2 cluster compose exec 'ls -a'` as an abbreviated, useless example).

janhenrikoverland:
Centralized and reuse e.g. commands -> cacheLocation?

amcgee:
> Centralize and reuse e.g. commands -> cacheLocation?

Yeah, definitely agree.  I'll leave that for later though since it would touch a bunch more files.  Thanks for the review @janhenrikoverland !

PR #100, opened 2019-07-15
chore: prepare 2.0 of d2

varl:
**BREAKING CHANGE:** Introduces cli-style v4.1.0

See https://github.com/dhis2/cli-style/blob/master/CHANGELOG.md#breaking-changes for a list of changes.


amcgee:
Wouldn't this be a breaking change technically?

varl:
> Wouldn't this be a breaking change technically?

We discussed this outside of the PR and while there is no technical reason to be strict about it at the moment, it does make sense to stay semantically correct and bump the major version when a sub-command changes drastically.

We might also consider moving d2-style into the cli repository at some point as well.

varl:
Make sure to squash with the commit message of:

> BREAKING CHANGE: Remove deprecated `d2-cluster seed` command, and update `d2-style` to 4.1.0.
>
> - For changelog of what has changed with cli-style, see https://github.com/dhis2/cli-style/blob/master/CHANGELOG.md#breaking-changes for a list of changes.
>
> - `d2-cluster seed` has been deprecated in favour of `d2-cluster db restore`, see `d2 cluster db --help` for more information on usage.

PR #101, opened 2019-07-17
feat: include the docs folder when pushing back a release commit

varl:
For libraries such as `ui-core`, `d2`, `prop-types`, it makes sense to generate a `docs/` folder when the release commit is built on CI, which is then pushed back to the default branch when a release commit is cut.

This would allow us to drop the additional step of pushing libraries to the d2-ci organisation on Github, and having docs together with the release.

### DIFF
--- a/packages/utils/src/cmds/release.js
+++ b/packages/utils/src/cmds/release.js
@@ -64,6 +64,7 @@ const handler = async ({ publish }) => {
         {
             assets: [
                 'CHANGELOG.md',
+                'docs',
                 packages.map(pkgJsonPath =>
                     path.relative(process.cwd(), pkgJsonPath)
                 ),
@@ -116,9 +117,7 @@ const handler = async ({ publish }) => {
             const { lastRelease, commits, nextRelease, releases } = result
 
             reporter.info(
-                `Published ${nextRelease.type} release version ${
-                    nextRelease.version
-                } containing ${commits.length} commits.`
+                `Published ${nextRelease.type} release version ${nextRelease.version} containing ${commits.length} commits.`
             )
 
             if (lastRelease.version) {
@@ -127,9 +126,7 @@ const handler = async ({ publish }) => {
 
             for (const release of releases) {
                 reporter.info(
-                    `The release was published with plugin "${
-                        release.pluginName
-                    }".`
+                    `The release was published with plugin "${release.pluginName}".`
                 )
             }
         } else {


packages/utils/src/cmds/release.js
```diff
@@ -116,9 +117,7 @@ const handler = async ({ publish }) => {
             const { lastRelease, commits, nextRelease, releases } = result
 
             reporter.info(
-                `Published ${nextRelease.type} release version ${
-                    nextRelease.version
-                } containing ${commits.length} commits.`
+                `Published ${nextRelease.type} release version ${nextRelease.version} containing ${commits.length} commits.`
```
varl: Prettier changed behaviour on how it splits template literals.

amcgee: 👍 I like this way better

amcgee:
Definitely in favor of docs living with the code, but I don't know that the **built** docs (i.e. static storybook site) should be pushed back to the **source** repo....

Is the idea that we would host the docs on netlify from the source repo instead of github-pages from the `d2-ci` repo?  If that's the case, shouldn't we "build" the docs within netlify (which we [already do](https://dhis2-ui.netlify.com/) for i.e. [ui-core](https://github.com/dhis2/ui-core/blob/master/netlify.toml)) and then we don't need to add anything here and can already remove the `d2-ci` push step?

amcgee:
Also, are we considering **storybook === docs** here?  Because that might not always be the case

varl:
> Also, are we considering **storybook === docs** here? Because that might not always be the case

The `docs/` folder may contain any static assets that can be deployed with GH Pages.

E.g. for **d2** and **prop-types** it is going to be a JSDoc site. For DHIS2 Apps it might be a Gitbook for User Documentation (in the future). For **ui-core** and **ui-widgets** it would be a static Storybook site.

> Is the idea that we would host the docs on netlify from the source repo instead of github-pages from the d2-ci repo?

No, I do not plan to host docs on Netlify. I want to move the docs on **d2-ci** to **dhis2** for libraries. We are using Netlify for some repositories now, but I still have concerns related to Netlify:

1. Only one person has admin rights over the Netlify integration
2. Even if we apply for a Open Source license from Netlify and we get a Team there, in exchange they [require us to put a link to Netlify on our main page at www.dhis2.org](https://www.netlify.com/legal/open-source-policy/)

**Number 1** is a problem because of the "hit by a bus" factor.

**Number 2** is a discussion that needs to be lifted with the management team if it is OK for DHIS2 to publicly endorse Netlify. What we use internally is one thing, but public endorsement of a for-profit service requires serious consideration.

> Definitely in favor of docs living with the code, but I don't know that the built docs (i.e. static storybook site) should be pushed back to the source repo....

I consider the generated docs as part of the release artifact, like the changelog that we generate and push back together with the release commit.

Each release commit on the release branch would have a changelog and the relevant docs attached to it.

varl:
>The source already contains the source for all the docs, just not the raw HTML that it eventually becomes.

I agree with this principle in general.

amcgee:
> * Only one person has admin rights over the Netlify integration
> 
> * Even if we apply for a Open Source license from Netlify and we get a Team there, in exchange they [require us to put a link to Netlify on our main page at www.dhis2.org](https://www.netlify.com/legal/open-source-policy/)

Option 3) Pay $15/month for additional users with no co-branding requirements?

amcgee:
> > * Only one person has admin rights over the Netlify integration
> > * Even if we apply for a Open Source license from Netlify and we get a Team there, in exchange they [require us to put a link to Netlify on our main page at www.dhis2.org](https://www.netlify.com/legal/open-source-policy/)
> 
> Option 3) Pay $15/month for additional users with no co-branding requirements?

They also may very well be OK with us putting the badges on our repositories

amcgee:
> Edit: A release commit is the generated result of a raw state, and consists of compiled information based on the Git history.

True, but I think we want to minimize the surface area of "generated" content in our repositories.  Right now it's fairly limited to `package.json` version and `CHANGELOG` file, but if an entire website under the `/docs` folder is also hanging around the temptation will be to manually edit that directory.

amcgee:
> There are some alternatives which I think are inferior:
> 
>     * We could generate docs when a PR is built on Travis, and push that back to the PR branch -- but that makes the code review process noisy.
> 
>     * We could publish the generated docs to the `gh-pages` branch so it's in sync with the release branch, but that still decouples the docs from the release they are tied to, a slight improvement over today but a more complicated flow in Travis.

I agree those are inferior options.

amcgee:
I'm not **super** opposed to this, but I will play devil's advocate because I think it introduces some additional complexity and confusion for developers looking at repositories.

varl:
> They also may very well be OK with us putting the badges on our repositories

I will explore that option.

> if an entire website under the /docs folder is also hanging around the temptation will be to manually edit that directory.

I'm not that worried about that:

- A PR would show that funky diff.
- It would be in `.gitignore`, and it should be overridden it with `git add --force docs` when pushed to the release commit.
- Even if merged, it would get overwritten on the next release.

> I think it introduces some additional complexity and confusion for developers looking at repositories.

This is a good point. 

We have some complexity today which is hidden and yet has caused problems for developers when setting up new libraries, or modifying the Travis configuration for existing apps/libs related to the obscured **d2-ci** organisation. 

Thinking on it, I am not sure if moving it from one isolated place that devs rarely interact with to a public place they regularly interact with is the right move. We won't get rid of the shadow organisation in the first place either way.

varl:
I checked the build sizes and the Storybook for **ui-core** is a hefty 7.6 MB and JSDocs for **d2** is a comparatively slim 3.2 MB. That's no fun.

varl:
> I will explore that option.

I talked within DHIS2 about if/where we can put the link, and with Netlify if we comply with the license if we do that, and looks like we would! 🎉

Talking to Netlify now about setting up an account under the open source license for DHIS2.

PR #113, opened 2019-08-15
feat: add i18n modernize to create new translation files from old ones

Mohammer5:
I'm using this one in the new maintenance app to create the translation files (`/i18n/*`) from the translation files in the old maintenance app (`/src/i18n/*`).

I'm not sure if this is the right place, but I guess it's a utility?
Also not sure about the name of the script/command.


packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+                        shouldAppendContents ? 'Appending' : 'Writing'
+                    } to "${newLanguageFilePath}"`
+                )
+            fs.writeFileSync(newLanguageFilePath, newContents, {
+                flag: shouldAppendContents ? 'a' : 'w',
+            })
+        }
+    }
+
+    argv.verbose && log.debug('Creating translation files :: Done')
```
amcgee: `reporter.debug` already handles only showing debug logs when `--verbose` is passed, so you don't need to use `argv.verbose && ...` everywhere

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+                newContents +=
+                    language === argv.primaryLanguage
+                        ? getTemplateMainLanguage()
+                        : getTemplateAlternativeLanguage(argv.appName, language)
+            }
+
+            for (const key in languageTranslations) {
+                if (languageTranslations.hasOwnProperty(key)) {
+                    if (!translations[argv.primaryLanguage][key]) {
+                        argv.logMissingKeys &&
```
amcgee: I don't think it should be necessary to have a specific `--log-missing-keys` option - if we use `log.debug` it only shows up when verbose anyway.  If we want to make them show up when **not** verbose, then maybe it makes sense to have an option.  Is there a reason you **wouldn't** want to know about missing keys?

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+                fs.existsSync(newLanguageFilePath) &&
+                argv.appendToExistingFiles
+
+            if (
+                !argv.overrideExistingFiles &&
+                fs.existsSync(newLanguageFilePath) &&
+                !argv.appendToExistingFiles
+            ) {
+                console.log('')
+                log.debug(
```
amcgee: Many of these `log.debug` calls should probably be `log.info` (for highlighted things like indicating steps in a process) or `log.print` (for lower-priority but still not verbose things)

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+        default: TRANSLATION_OUT_DIR,
+    },
+
+    inDir: {
+        describe: 'Input directory where old translation files can be found',
+        type: 'string',
+        default: TRANSLATION_IN_DIR,
+    },
+
+    createOutDir: {
```
amcgee: Why would you not want to create the output directory if it doesn't exist?

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+        default: 'true',
+    },
+
+    overrideExistingFiles: {
+        describe: 'Overriding the contents of existing translation files',
+        type: 'boolean',
+        default: 'false',
+    },
+
+    appendToExistingFiles: {
```
amcgee: Should this default to true?

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+        describe:
+            'Appends the new contents to existing translation files, can only be used when not using `--override-existing-files`',
```
amcgee: Add `conflicts: 'overrideExistingFiles'` to let `yargs` enforce that this can't be used when `overrideExistingFiles` is also supplied.

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+    },
+}
+exports.handler = argv => {
+    const inDirExists = fs.existsSync(argv.inDir)
+    const outDirExists = fs.existsSync(argv.outDir)
+    const languagesToTransform = argv.languages
+        ? argv.languages.split(/,\s*/)
+        : []
+
+    if (!argv.appName) {
```
amcgee: Instead of checking for required args manually, pass `required: true` to the option builder.

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+                argv.inDir,
+                mainTranslationFile
+            )}")`
+        )
+        process.exit(1)
+    }
+
+    argv.verbose &&
+        log.debug('Extracting key/value pairs from translation files')
+    const translations = translationFiles.reduce((mappings, file) => {
```
amcgee: It would be nice to factor this function a bit, could we move everything that's not argument parsing logic out into a separate file and separate out, for instance, the body of this reduce function so everything isn't inline here?

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+                argv.outDir,
+                newLanguageFileName
+            )
+            const shouldAppendContents =
+                !argv.overrideExistingFiles &&
+                fs.existsSync(newLanguageFilePath) &&
+                argv.appendToExistingFiles
+
+            if (
+                !argv.overrideExistingFiles &&
```
amcgee: This is duplicated above, I think you meant to use the `shouldAppendContents` variable?

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+"Project-Id-Version: i18next-conv\\n"
+"MIME-Version: 1.0\\n"
+"Content-Type: text/plain; charset=utf-8\\n"
+"Content-Transfer-Encoding: 8bit\\n"
+"Plural-Forms: nplurals=2; plural=(n != 1)\\n"
+"POT-Creation-Date: 2019-08-13T13:38:22.134Z\\n"
+"PO-Revision-Date: 2019-08-13T13:38:22.134Z\\n"
+`
+
+const getTemplateAlternativeLanguage = (appName, language) => `
```
amcgee: It would be nice to extract these templates to different files as well, you could just keep them as raw files and read them from the filesystem with something like `fs.readFileSync(path.join(__dirname, 'templatfile.txt'))`

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+ *   --app-name maintenance-app \
+ *   --languages fr,ar \
+ *   --delete-old-files \
+ *   --log-missing-keys
+ */
+const path = require('path')
+const fs = require('fs')
+const log = require('@dhis2/cli-helpers-engine').reporter
+
+const CONSUMING_ROOT = path.join(process.cwd())
```
amcgee: It might be worth allowing the cwd to be passed in as an argument?

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+"Content-Transfer-Encoding: 8bit\\n"
+"X-Generator: Translate Toolkit 2.2.5\\n"
+"X-Pootle-Path: /${language}/${appName}/${language}.po\\n"
+"X-Pootle-Revision: 29879\n"
+`
+
+exports.command = 'transform_translate'
+exports.describe = 'Transform old translation file style to new style'
+exports.builder = {
+    appName: {
```
amcgee: I think it makes sense to have some arguments, rather than having everything as options.  `appName`, `inDir`, and `outDir` seem like good candidates.  You can specify positional arguments (required and optional) in the `command` name like this: `transform_translate <appName> [inDir] [outDir]`, though it might make sense that both dirs be required and I'm not sure it's obvious to the user why `appName` should be...

amcgee:
This seems like a really handy utility, thanks for putting it together @Mohammer5 !

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
...
+        describe:
+            'The name of the app, must be lower case, use dashes instead of whitespaces and be prefixed with "-app". Normally this should the url part of the github repo (maintenance app => "maintenance-app")',
```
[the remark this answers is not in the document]
Mohammer5: > Side note: why are we enforcing that the names follow this format?

I'm not sure if we need to enforce it.
The `.po` files container a `"X-Pootle-Path: /${language}/${appName}/${language}.po\\n"` part, I have no idea what it's being used for right now. 

Will investigate!

packages/utils/src/cmds/transform_translate.js
Mohammer5: It can be a very high number of keys missing (that's what happened to me in the maintenance app), it rendered all other output useless..

I thought about adding a `--dry-run` option, which could be combined with `--log-missing-keys` and `--languages` to get a list of all missing keys without any debug output, what do you think?

packages/utils/src/cmds/transform_translate.js
Mohammer5: I think it's safer to abort the process before doing any harm,
automated modifying of files should be a conscious choice in my opinion

packages/utils/src/cmds/transform_translate.js
Mohammer5: No, this condition is true when `appendToExistingFiles` is `false`, while the variable above is only true when `appendToExistingFiles` is `true`.

I will move the variable to where it's being used though, this is not the right place (which caused the confusion in the first place I assume)

packages/utils/src/cmds/transform_translate.js
```diff
@@ -0,0 +1,381 @@
+/**
+ * Example usage:
+ *
+ * d2-utils transform_translate \
```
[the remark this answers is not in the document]
Mohammer5: I went for this one: `d2 app i18n modernize`

packages/utils/src/cmds/transform_translate.js
Mohammer5: That'd be redundant as you can do that already with `inDir` and `outDir`

packages/utils/src/cmds/transform_translate.js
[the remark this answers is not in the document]
Mohammer5: `appName` has been replaced with `pootlePath`, which is not optional

packages/utils/src/cmds/transform_translate.js
Mohammer5: @amcgee How do I use options when using the nested namespaces approach? 
Now the command is a key, not a string anymore

Mohammer5:
@amcgee I've addressed all of your concerns or left questions, could you please have another look?

amcgee:
Yeah, thanks @Mohammer5 I'll take a look soon!

packages/app/src/commands/modernize.js
```diff
@@ -0,0 +1,162 @@
+/**
+ * Example usage:
+ *
+ * d2-utils modernize \
```
amcgee: This would be great as a `jsdoc` comment to auto-generate documentation!  For now fine to keep it as a comment here :-)

(side-note: should be `d2-app i18n modernize` now)

packages/app/src/commands/modernize.js
```diff
@@ -0,0 +1,162 @@
...
+    if (deleteOldFiles) {
+        reporter.info('Deleting legacy files')
+        deleteLegacyFiles({
+            translationFiles,
+            languagesToTransform,
+        })
+    }
+}
+
+const modernize = {
```
amcgee: I think the issue here is that you needed a `command` key to tell `yargs` the name of the command.  I restructured it a little bit to match how we do namespacing other places, see #126

packages/app/src/commands/modernize.js
```diff
@@ -0,0 +1,162 @@
...
+    pootlePath,
+    deleteOldFiles,
+    logMissingKeys,
+    primaryLanguage,
+    overrideExistingFiles,
+    appendToExistingFiles,
+}) => {
+    const languagesToTransform = languages ? languages.split(/,\s*/) : []
+
+    const translationFiles = fs
```
amcgee: This throws a pretty nasty error if `inDir` doesn't exist.  We're checking for existence later in `checkRequirements` but those checks probably need to be moved here.

See https://github.com/dhis2/cli/pull/126/files#diff-f0dae67115941bb5d80b603159a41f59R111-R115

packages/app/src/commands/modernize/checkRequirements.js
```diff
@@ -0,0 +1,54 @@
...
+ * @param {string[]} args.translationFiles
+ * @return {void}
+ */
+const checkRequirements = ({
+    inDir,
+    outDir,
+    primaryLanguage,
+    translationFiles,
+}) => {
+    const inDirExists = fs.existsSync(inDir)
```
amcgee: These checks should probably also confirm that `inDir` is actually a directory, not a file.  You can do that with `fs.existsSync(inDir) && fs.statSync(inDir).isDirectory()` (see [here](https://github.com/dhis2/cli/pull/126/files#diff-f0dae67115941bb5d80b603159a41f59R111-R115))

packages/utils/src/cmds/transform_translate.js
amcgee: @Mohammer5 not sure I understand the question... I am able to see the help and pass options properly with `./bin/d2-app i18n --help`, `./bin/d2-app i18n modernize --help`, and `./bin/d2-app i18n modernize --inDir <something>` on this branch.  What do you mean by "the command is a key, not a string anymore"?

I also restructured the directories to match the structure we use elsewhere in #126 (feel free to merge that or just reference it) - keeping helper code outside the `commands` directory and having a separate file for each namespace (`i18n`) and command (`modernize`)

amcgee: Ah, my bad I see what you mean - when using the `commands` object to create the namespace the key becomes the command name.  I believe you can still pass the `command: 'modernize <appName> [inDir] [outDiir]'` in that case by placing them in the key or in the `command` property of the command object, or just avoid using `commands: { ... }` and use `builder: yargs => yargs.command(modernize)` instead

amcgee: Examples, tested and these all work:

```js
const modernize = {
    describe: 'Transform old translation file style to new style',
    builder,
    handler,
}

module.exports = namespace('i18n', {
    description: 'Handle translations in apps',
    commands: { 'modernize <appName>': modernize },
})
```

```js
const modernize = {
    command: 'modernize <appName>'
    describe: 'Transform old translation file style to new style',
    builder,
    handler,
}

module.exports = namespace('i18n', {
    description: 'Handle translations in apps',
    builder: yargs => yargs.command(modernize)
})
```

```js
//src/commands/i18n/modernize.js
module.exports = {
    command: 'modernize <appName>'
    describe: 'Transform old translation file style to new style',
    builder,
    handler,
}

// src/commands/i18n.js
module.exports = namespace('i18n', {
    description: 'Handle translations in apps',
    builder: yargs => yargs.commandDir('i18n')
})
```

packages/utils/src/cmds/transform_translate.js
amcgee: OK cool, I think either `--log-missing-keys` or `--dry-run` is fine, whichever you prefer.  We could eventually add a tool for analyzing the modern i18n and finding missing keys, but for now I don't think it's a crucial step of the i18n modernize process. 👍

Mohammer5:
@amcgee Thanks for your PR, I've changed things a little bit and merged it into this branch & adressed the options/parameter thing as well.

Mohammer5:
@amcgee could you have another look at this one?

packages/app/src/commands/i18n/modernize.js
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * npx -p @dhis2/cli d2-app i18n modernize \
+ *   ~/development/dhis2/maintenance-app/src/i18n # inDir \
+ *   ~/development/dhis2/project-doom/i18n # outDir \
+ *   --override-existing-files \
+ *   --pootle-path "/app/import-export-app/" \
+ *   --languages fr,ar \
+ *   --delete-old-files \
+ *   --log-missing-keys
+ */
```
amcgee: _(minor)_ this could just be `npx @dhis2/cli app i18n modernize` or `npx @dhis2/cli-app i18n modernize`, there shouldn't be a need for the `-p` or `d2-app`

packages/app/src/helpers/modernize/alternative_language.template
```diff
@@ -0,0 +1,15 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: PACKAGE VERSION\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: ###CREATION_DATE###\n"
+"PO-Revision-Date: ###REVISION_DATE###\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
```
amcgee: Should we just omit this since it's not really used?

packages/app/src/helpers/modernize/alternative_language.template
```diff
@@ -0,0 +1,15 @@
...
+"Language-Team: LANGUAGE <LL@li.org>\n"
```
amcgee: This seems wrong, should we just remove it?

jason-p-pickering: @amcgee pretty sure that is needed, since is part of the PO spec. That it is wrong ,is a totally different problem I think.

amcgee: @jason-p-pickering got it, thanks.  @Mohammer5 do we know where this li.org email came from?

amcgee: I don’t know much about the PO spec or the history here, so it’s possible this is fine as-is

Mohammer5: Looks like these are the default values:
https://www.gnu.org/software/trans-coord/manual/gnun/html_node/PO-Header.html

I think we could omit all the others and let these be handled manually:

```
"Report-Msgid-Bugs-To: \n"
"Language: ###LANGUAGE###\n"
"X-Generator: Translate Toolkit 2.2.5\n"
"X-Pootle-Path: ###POOTLE_PATH###\n"
"X-Pootle-Revision: 0\n"
```

<hr />

I'll add an argument for the `LANGUAGE` one, which will default to `LANGUAGE <LL@li.org>`

amcgee: `li.org` doesn't resolve, does the `LL@li.org` exist?  This header is suppose to specify "the mailing list on which the translation team can be reached" - seems like maybe we should put a DHIS2 email in there if anything.

In any case, since these files are actually parsed by our own tools and only used internal to the repo (and soon in transifex) I don't think we need most of these headers, unless we're trying to be backwards-compatible with Pootle.  I haven't found any references to these headers in the [Transifex](https://docs.transifex.com/formats/gettext) or [Gettext](https://www.gnu.org/software/gettext/manual/gettext.html#PO-Files) docs.

I vote either for removing these lines or leaving them as-is (wrong, but not super harmful) rather than adding options for them.

amcgee:
Hey @Mohammer5 !  I think this is probably fine with or without the few different defaults, we don't really use those headers anyway.  Is this ready to merge or did you want to make any changes?

Mohammer5:
> @Mohammer5 up to you if you want to make the above requested changes

Sorry, must've overlooked your previous comment.
I'll just merge it. I think it's ok

PR #124, opened 2019-08-29
ci: update travis tokens for npm and github

varl:


.travis.yml
```diff
@@ -1,23 +1,23 @@
 language: node_js
```
varl: The Travis CLI tool insists on this format.

PR #146, opened 2019-09-27
feat: add script that sets up cypress and cucumber

Mohammer5:
- [x] Install cypress
- [x] Install cypress-cucumber-preprocessor
- [x] Add folders & config files
- [x] Add DHIS 2 specific cypress helpers
- [x] Add Test feature
- [x] add `cypress.env.json` to .gitignore
- [x] Create script that will run both the local app and cypress
  - [x] Ask for required env var values if not present in `cypress.env.json` and either add to or create file
- [ ] Maybe refactoring
- [ ] Maybe with storybook as frontend / without backend authentication (for e. g. `ui-core` or `ui-forms`)


Mohammer5:
Safe to close this one as it'l be part of its own repo that handles cypress in the future.

PR #161, opened 2019-10-10
feat: allow defer-release keyword in commit messages

varl:


amcgee:
OPTIONAL - should we be a little more lax about the format?  Something like `message.toLowerCase().match(/[\s*defer[ -]release\s*]/g)` could work, but maybe it's fine to leave it strict for now!

varl:
>`message.toLowerCase().match(/[\s*defer[ -]release\s*]/g)`

It is an insensitive match (`i`), so there's some slack in what it accepts. I think the brackets should be escaped though, otherwise it lists the characters that it accepts. Not sure why we'd accept whitespace before/after the tag, but I do like the dash or space.

Will changed to `/\[defer[ -]release\]/gi` so these would match:

```
[defer release]
[defer-release]
[Defer Release]
```

PR #216, opened 2019-12-02
fix(release): avoid a stack trace when no commits are passed

varl:
Ours isn't the only plugin that crashes here, but we can give some more information about what goes wrong.


varl:
Fixed by https://github.com/dhis2/cli/pull/214

PR #236, opened 2020-01-09
feat(cluster): add option to pass in path to custom dhis.conf

varl:
Requires https://github.com/dhis2/docker-compose/pull/10

### DIFF
--- a/.huskyrc.js
+++ b/.huskyrc.js
@@ -9,5 +9,6 @@ module.exports = {
             'd2-style js check --staged',
             'd2-style text check --staged',
         ]),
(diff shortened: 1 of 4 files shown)


amcgee:
Nice, but am I missing something?  How does this actually pass the file to the Docker container?

varl:
> Nice, but am I missing something? How does this actually pass the file to the Docker container?

It sets the `DHIS2_CORE_CONFIG` variable on the env, which is passed to Docker Compose when we run it.

Then the variable substitution in dhis2/docker-compose#10 takes over and uses it if it is set, and falls back to the default if not.

PR #242, opened 2020-01-10
docs: add docsite and initial structure

varl:


varl:
@amcgee this adds the netlify.toml so I'll set it up right after merging.

PR #385, opened 2020-11-10
fix(schema-diff): convert properties-array to object for better diffs

Birkbjo:
Previously it was hard to know which property a change in the diff was related to, as it only showed the index.

Now we are transforming the properties-array to an object hashed by the `name` of the property. This is implemented pretty generically, so it should work for eg. the `authorities` array was well, but the most important change is for the `properties` array.

### DIFF
--- a/packages/utils/src/cmds/schema/diff/index.js
+++ b/packages/utils/src/cmds/schema/diff/index.js
@@ -15,10 +15,18 @@ const {
 } = require('../')
 
 let cache
+
+const hashKeys = ['singular', 'name', 'fieldName', 'type']
 // We use the singular property as an unique identifier for schemas
 // name and type are used for other nested properties
 const objectHash = obj =>
     obj.singular || obj.name || obj.fieldName || obj.type || obj
+
+const getHashKeyForObject = obj =>
+    hashKeys.find(hashKey => Object.keys(obj).find(prop => hashKey === prop))
+
+const getHashKeyForArray = array => getHashKeyForObject(array[0])
+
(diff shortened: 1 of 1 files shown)


varl (approved):
LGTM!

packages/utils/src/cmds/schema/diff/index.js
```diff
@@ -81,18 +89,50 @@ function sortArrayProps(obj) {
...
+    return array.reduce((acc, curr, index) => {
+        let hash = curr[hashKey]
+        if (!hash) {
+            reporter.warn(
+                'hash-candidate value not found, falling back to index for ',
+                curr
+            )
+            hash = index
+        }
+        acc[curr[hashKey]] = curr
```
amcgee, line 102: I think this needs to be `acc[hash]` in order to follow the logic in the if statement (if the hash candidate isn't found) - otherwise `curr[hashKey]` will be undefined here and so the different array values will overwrite eachother

amcgee (changes requested):

packages/utils/src/cmds/schema/diff/index.js
varl: cc: @Birkbjo

amcgee: @Birkbjo pinging this one again

Birkbjo: Sorry for the super late reply. This has drowned in my inbox. But you are right, will fix.

PR #393, opened 2020-12-01
chore: updates `d2-utils-release` documentation

paschalidi:
Hey this updates the `release` part of the documentation a little bit. Please have a look and let me know if there is anything you would phrase differently.

This now looks like this 
image


docs/commands/d2-utils-release.md
```diff
@@ -12,6 +12,19 @@ understanding how that tool operates is helpful.
...
+In detail when a commit that utilizes semantic release appears on `master` the following will happen:
+
+1. A Github release similar to [this one](https://github.com/dhis2/usage-analytics-app/releases)
+   will be published.
+
+2. The `CHANGELOG.md` file will be updated. An example can be found
+   [here](https://github.com/dhis2/usage-analytics-app/blob/master/CHANGELOG.md).
+   _The first release will automatically create a `CHANGELOG.md` in the root directory of your repository._
+
+3. The versions in the `package.json` will be updated as well.
```
Mohammer5, line 24: This does also apply to yarn monorepos. The release command will update all the versions of the workspaces' package.json files to the same, new version.

Mohammer5 (changes requested):
I think we should mention that the package.json needs to contain the following in case the module should be published on npm:

```json
{
  "private": false,
  "publishConfig": {
    "access": "public"
  }
}
```

paschalidi:
@Mohammer5 thanks for looking into it. I have applied what you suggested. This is how this will look now 👇   
image

Mohammer5 (approved):
LGTM

paschalidi:
So now the question is why these jobs run for ever. @varl any ideas?

PR #439, opened 2021-05-13
docs: clarify documentation of d2 cluster down --clean

deboragaleano:
This PR adds suggestion to clarify documentation of the command `d2 cluster down --clean`. Jira issue [CLI-26](https://jira.dhis2.org/browse/CLI-26)
Please let me know if there is anything that should be specified to make this more clear.


docs/commands/d2-cluster.md
```diff
@@ -87,10 +87,34 @@ command to be able to do anything at all: `{command}` and `{name}`.
...
+
+This command destroys a running container: 
+
+```bash
+d2 cluster down {name}
+```
+
+#### Command `down --clean`
+
+This command destroys all data volumes as well as ephemeral containers:
```
varl, line 112:
```suggestion
This command brings down a cluster and cleans up after itself. This destroys all containers and volumes associated with the cluster. For example, this means that the attached database will be wiped so it is useful when you want to remove a cluster entirely.
```

varl (approved):
[MINOR] We can be more clear in the difference between `down` and `down --clean` I think and give the reader an idea of when to use it.

PR #514, opened 2021-11-15
feat(cli): prefer project-local tools

varl:
Increases the utility of the global d2 tool for developers by preferring
the version of the CLI tool that is installed over the globally bundled
one.

Before this change these two commands run in the same project can yield
different results:

    d2 style check          # global
    yarn d2-style check     # local

After the change, they will both use the local version:

    d2 style check          # local
    yarn d2-style check     # local

Jira-ticket-key: CLI-68
Jira-ticket-URL: https://jira.dhis2.org/browse/CLI-68

Depends on:
- https://github.com/dhis2/cli-helpers-engine/pull/143

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -16,7 +16,7 @@
     "private": false,
     "dependencies": {
         "@dhis2/cli-app-scripts": "^8.3.0",
-        "@dhis2/cli-helpers-engine": "^3.0.0"
+        "@dhis2/cli-helpers-engine": "^3.1.0"
     },
     "publishConfig": {
         "access": "public"
--- a/packages/app/src/index.js
+++ b/packages/app/src/index.js
@@ -1,9 +1,13 @@
-const { namespace } = require('@dhis2/cli-helpers-engine')
+const { namespace, loadModule } = require('@dhis2/cli-helpers-engine')
 
 module.exports = namespace('app', {
     description: 'Front-end application and library commands',
     builder: yargs => {
-        yargs.command(require('@dhis2/cli-app-scripts'))
+        const loader = loadModule({
+            parentModule: __filename,
+        })
+
+        yargs.command(loader('@dhis2/cli-app-scripts'))
         yargs.commandDir('commands')
     },
 })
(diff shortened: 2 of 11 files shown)


amcgee (changes requested):
Implementation looks great! Can we move this to `cli-helpers-engine` though and then use it to load all sub-modules?  For example, this will find a local version of `d2-app` but not `d2-app-scripts` which would be very handy

amcgee:
Also, should we maybe have a flag to short-circuit the lookup and revert to the global version?

mediremi (approved):

varl:
> Also, should we maybe have a flag to short-circuit the lookup and revert to the global version?

I was thinking of adding `--global` to do that, or `--prefer-global` with a short alias (`-g`, probably). 

I had one variant that did this but I think we'd need to manually parse `yargs.argv` (as it's not final) and pass it from the builder to the `load` function. I can see if the situation changes when we move it to engine.

varl:
> Implementation looks great! Can we move this to `cli-helpers-engine` though and then use it to load all sub-modules? For example, this will find a local version of `d2-app` but not `d2-app-scripts` which would be very handy

Well, that was quick !

Sure, I can move it. :+1:

varl:
> Also, should we maybe have a flag to short-circuit the lookup and revert to the global version?

This happens in the builder phase, before the arguments have been parsed. Moving the code later in `makeEntrypoint` causes problems with yargs, and calling `.argv` / `.parse()` in the builder functions is [not recommended](https://github.com/yargs/yargs/blob/main/docs/api.md#argv):

> **Note**: `.argv` should only be used at the top level, not inside a command's builder function.

And I can verify that it is indeed problematic.

The `loadModule` function accepts a boolean to toggle `preferLocal` on/off, so the functionality is there, and would be great if we could get it from yargs directly as it would be very easy to integrate:

```js
const globalLoader = loadModule({
    parentModule: __filename,
    preferLocal: yargs.argv.local,
})
```

varl:
@amcgee @mediremi Updated to use the CLI Engine function.

mediremi (approved):

varl:
@amcgee could you re-review this ? Currently blocked by the old review.

amcgee (approved):
Awesome 😎

PR #565, opened 2022-09-15
feat(cluster): context path

Birkbjo:
Previously the `--customContext` option for the `up`-command was just a boolean, and it used the `cluster name` when this was set. I think it should be possible to set this independently of the `cluster name`, as it's very useful if you would wanted additional levels, eg. `/dhis2/dev`. This is somewhat related to [CLI-75](https://dhis2.atlassian.net/browse/CLI-75), since having a cluster with a path-delimeter (/), results in very buggy behavior.


packages/cluster/src/commands/list.js
```diff
@@ -36,7 +36,9 @@ const formatStatus = status => {
 
 const run = async function (argv) {
     const clusters = await listClusters(argv)
-
+    const anyCustomContext = clusters.some(
+        cluster => cluster.contextPath !== ''
+    )
```
amcgee, line 41: Should we just always show the context, even if none are custom?

packages/cluster/src/commands/list.js
```diff
@@ -45,20 +47,22 @@ const run = async function (argv) {
...
-            ])
+            table.push(
+                [
+                    chalk.blue(cluster.name),
+                    cluster.port,
+                    cluster.channel,
+                    cluster.dhis2Version,
+                    cluster.dbVersion,
+                    formatStatus(status),
+                ].concat(anyCustomContext ? cluster.contextPath : [])
```
amcgee, line 64: Maybe `cluster.contextPath || '/'` ?

amcgee (approved):
Haven't tested myself but this looks good, thanks @Birkbjo!  Some minor comments, not blocking.

packages/cluster/src/commands/list.js
Birkbjo: We could. The idea was that this is most likely not relevant to most users. I think very few are actually using this context-path, and it might be confusing if you don't know what it does?

packages/cluster/src/commands/list.js
Birkbjo: If you think we should show this regardless, I agree, we should probably do that.

packages/cluster/src/commands/list.js
amcgee: I could go either way here... I also think it would be confusing if `d2 cluster list` sometimes renders in a different format if you happen to have a custom context on one cluster you're running.  It would make any kind of parsing of the output difficult as well, though we're not exactly well setup for that as it is

PR #586, opened 2022-12-21
fix: set DHIS2_HOME env var to /DHIS2_home for images pre-Jib

teleivo:
Docker images created for releases after `2.39.0`, `2.38.2`, `2.37.8.1` do not create directory `/DHIS2_home` anymore. See https://github.com/dhis2/dhis2-core/pull/11981
Instead `/opt/dhis2` is created which is the default location used by DHIS2.

Users of the new Docker image cannot simply mount a directory to `/DHIS2_home` and set env var `DHIS2_HOME=/DHIS2_home` as the mounted directory will by default not be writable by others. Docker gives the directory the same uid, guid of the host which will not exist in the container. A user would need to make the directory writable by others on their host. This is needed as the directory will not be owned by the user running DHIS2 (Tomcat) and DHIS2 requires write permission to DHIS2_HOME.


radnov (approved):

packages/cluster/src/defaults.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+    dhis2Home: '/opt/dhis2',
```
amcgee, line 2: (minor) As far as I can tell this will never be used because it is always overridden by the "computed" dhis2home (common.js::168), is that correct and if so should this default be removed...?

teleivo: I think you are right 👌🏻 I thought I add it here to be consistent since others like `dhis2Version` also have a "default" even though for `dhis2Version` the default is `''`.

Keeping this default or removing is both fine for me. Which do you prefer @amcgee @radnov ?

radnov: Given that `dhis2Version` is also in the defaults with an empty value, maybe we can either keep it as `/opt/dhis2` (if we assume this is the default home location for images going forward) or also keep it with an empty value. I'm also fine either way. 🤷‍♂️  😅

Philip-Larsen-Donnelly (approved):

PR #626, opened 2024-08-14
fix: move to docker compose v2

ghost:
BREAKING CHANGE: drop docker compose v1 compatibility, as that's no longer maintained. See: https://docs.docker.com/compose/migrate/


KaiVandivier (approved):
I'm not too familiar with this code, but it looks good and I started up a server using it 🙂

PR #641, opened 2026-02-12
feat(create-app): add custom template support via GitHub template sources 

derrick-nuby:
### Summary

This PR adds support for using a GitHub repository as a template source in `@dhis2/create-app`.
In addition to built-in templates (`basic`, `react-router`), users can now scaffold from public GitHub template repositories using shorthand or URL formats, with optional branch targeting.

### What Changed

1. Added template source parsing and detection for GitHub specifiers.
2. Added external template source resolver for remote GitHub templates cloned into a temporary directory.
3. Updated create flow to keep built-in template resolution in `index.js`, and use external resolver + cleanup in `finally` for Git templates.
4. Added interactive flow option for `Custom template from Git`.
5. Updated create-app docs and command docs with usage examples.

### Supported Template Formats

1. `owner/repo`
2. `owner/repo#ref`
3. `https://github.com/owner/repo`
4. `https://github.com/owner/repo#ref`

### Validation

Executed local and smoke tests for:

1. Existing built-in template flow (regression-safe behavior).
2. GitHub URL template source.
3. GitHub URL with branch ref.
4. Shorthand `owner/repo#ref`.

### Notes

1. Scope intentionally focuses on GitHub and public repositories.
2. Private repo auth UX and non-GitHub hosts are out of scope for this PR.
3. I did not include tests as this repo doesn't have a test suite, but I manually verified all scenarios and flows.


packages/create-app/src/index.js
```diff
@@ -158,8 +172,27 @@ const command = {
         }
 
         reporter.info('Copying template files')
-        const templateFiles = getTemplateDirectory(selectedOptions.templateName)
-        fs.copySync(templateFiles, cwd)
+        const builtInTemplateMap = {
```
kabaros: maybe for built-in templates, just resolve here in this module - no need to create a map, pass it and resolve it .. the `getTemplateDirectory` (renamed to `getBuiltInTemplateDirectory`) already does most of that responsibility.

So something like 
```js
let resolvedTemplate = getBuiltInTemplateDirectory(selectedOptions.templateSource)

if(!resolvedTemplate) {
  // then do everything related to Git repos including cleanup
 // 
}
```

no need to pass the map to resolveTemplateSource or make it aware of built-in tempaltes. (maybe rename the module to `resolveExternalTemplates`)

packages/create-app/src/utils/resolveTemplateSource.js
```diff
@@ -0,0 +1,127 @@
+const os = require('node:os')
+const path = require('node:path')
+const { exec } = require('@dhis2/cli-helpers-engine')
+const fs = require('fs-extra')
+const {
+    isGitTemplateSpecifier,
+    parseGitTemplateSpecifier,
+} = require('./isGitTemplateSpecifier')
+
+const ensureTemplateDirectory = (templatePath, templateSource) => {
```
kabaros: maybe better named `validateTemplateDirectory`? 

- also could you move these helper methods into their own files (similar to `isGitTemplateSource`) .. a lot if happening in this file now

kabaros (commented):
Thanks a lot for this @derrick-nuby ! I left some comments here .. let me know if it's not clear, or you disagree

Thanks again for the great code, and taking initiative in adding this feature 🙏🏿

kabaros (approved):

